PR #1, opened 2017-04-11
Repo structure and k8s client

rossf7:
Towards giantswarm/giantswarm#1086

- Repo structure with readme, license etc
- Small initial scope with util for creating Kubernetes Clients


glide.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - error
+  - logger
+- package: k8s.io/client-go
+  version: v2.0.0
+  subpackages:
+  - kubernetes
+  - rest
+testImport:
+- package: github.com/stretchr/testify
+  version: ^1.1.4
```
JosephSalisbury: Random drive by, let's pin this

client/k8s/k8s.go
```diff
@@ -0,0 +1,101 @@
...
+	var err error
+
+	if config.InCluster {
+		config.Logger.Log("debug", "creating in-cluster config")
+		rawClientConfig, err = rest.InClusterConfig()
+		if err != nil {
+			return nil, microerror.MaskAny(err)
+		}
+
+		if config.Address != "" {
```
kopiczko: Does it make sense to use external API for in-cluster operator?

rossf7: Good point, I've removed.

client/k8s/k8s.go
```diff
@@ -0,0 +1,95 @@
...
+func getRawClientConfig(config Config) (*rest.Config, error) {
+	var rawClientConfig *rest.Config
+	var err error
+
+	if config.InCluster {
+		config.Logger.Log("debug", "creating in-cluster config")
+		rawClientConfig, err = rest.InClusterConfig()
+		if err != nil {
+			return nil, microerror.MaskAny(err)
+		}
```
kopiczko: Maybe a nitpicking, but may be replaced with `return rawClientConfig, microerror.MaskAny(err)`. Then `else` is not needed.

kopiczko: Also `rawClientConfig` and `err` won't need to be declared on top.

rossf7:
@xh3b4sd coool. There is a WIP PR in cert-operator showing it being used. It needs a rebase as its pretty old but I'll get that sorted.

https://github.com/giantswarm/cert-operator/pull/28/files

kopiczko:
wip

Issue #13, opened 2017-05-24
Set resync period in operatorkit


rossf7:
See https://github.com/giantswarm/cert-operator/issues/49

- [x] added to `operatorkit` https://github.com/giantswarm/operatorkit/pull/14
- [x] used in `aws-operator` https://github.com/giantswarm/aws-operator/pull/275
- [x] used in `cert-operator` https://github.com/giantswarm/cert-operator/pull/55

PR #14, opened 2017-05-24
add ResyncPeriod constant

asymmetric:
used to determine the interval at which the lister function is called

Towards #13.

### DIFF
--- a/tpr/tpr.go
+++ b/tpr/tpr.go
@@ -13,6 +13,10 @@ import (
 )
 
 const (
+	// ResyncPeriod is the interval at which the Informer cache is invalidated,
+	// and the lister function is called.
+	ResyncPeriod = 1 * time.Minute
+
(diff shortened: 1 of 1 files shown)


tpr/tpr.go
```diff
@@ -14,6 +14,10 @@ import (
 
 const (
 	tprInitMaxElapsedTime = 2 * time.Minute
+
+	// resyncPeriod is the interval at which the operator cache is invalidated,
+	// and the lister function is called.
+	resyncPeriod = 1 * time.Minute
```
kopiczko: I think it should be capitalized :)

BTW compilation should fail I think and CI passes...

kopiczko: I was wrong, go compiler is OK about unused constants.

asymmetric: True story, I capitalized it and now it's there for you to review :)

PR #15, opened 2017-05-25
Infer TPR kind and group

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/1483

This change allows passing full name of TPR and extract kind name and group name from it. This is useful for creating TPOs (used in kubernetesd), and also allows creating REST endpoints (used in all operators, and potentially in reconciliation loop).

Next steps is to add pluralization logic and provide methods with TPR resource endpoint and TPR watch endpoint.

### DIFF
--- a/tpr/kind.go
+++ b/tpr/kind.go
@@ -11,7 +11,7 @@ import (
 	https://github.com/kubernetes/kubernetes/blob/6a4d5cd7cc58e28c20ca133dab7b0e9e56192fe3/pkg/registry/extensions/thirdpartyresourcedata/util.go
 */
 
-// extractKindAndGroup extracts kind and group from a name. For details see
+// extractKindAndGroup extracts kind and group from a name. For details see the
 // Config.Name godoc.
 func extractKindAndGroup(name string) (kind, group string, err error) {
 	parts := strings.Split(name, ".")
@@ -22,8 +22,8 @@ func extractKindAndGroup(name string) (kind, group string, err error) {
 	// kind
 	kindPart := parts[0]
 	toUpper := true
-	for ix := range kindPart {
-		char := kindPart[ix]
+	for i := range kindPart {
+		char := kindPart[i]
 		if toUpper {
 			kind = kind + string([]byte{(char - 32)})
 			toUpper = false
--- a/tpr/kind.go
+++ b/tpr/kind.go
@@ -13,7 +13,7 @@ import (
 
 // extractKindAndGroup extracts kind and group from a name. For details see
 // Config.Name godoc.
-func extractKindAndGroup(name string) (kind string, group string, err error) {
+func extractKindAndGroup(name string) (kind, group string, err error) {
 	parts := strings.Split(name, ".")
 	if len(parts) < 3 {
 		return "", "", microerror.MaskAnyf(unexpectedlyShortResourceNameError, "%s, expected at least <kind>.<domain>.<tld>", name)
--- a/tpr/kind.go
+++ b/tpr/kind.go
@@ -1,0 +1,40 @@
+package tpr
+
+import (
+	"strings"
+
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+/*
+	Code based on:
+	https://github.com/kubernetes/kubernetes/blob/6a4d5cd7cc58e28c20ca133dab7b0e9e56192fe3/pkg/registry/extensions/thirdpartyresourcedata/util.go
+*/
+
+// extractKindAndGroup extracts kind and group from a name. For details see
+// Config.Name godoc.
+func extractKindAndGroup(name string) (kind string, group string, err error) {
+	parts := strings.Split(name, ".")
+	if len(parts) < 3 {
+		return "", "", microerror.MaskAnyf(unexpectedlyShortResourceNameError, "%s, expected at least <kind>.<domain>.<tld>", name)
+	}
+
+	// kind
+	kindPart := parts[0]
+	toUpper := true
+	for ix := range kindPart {
+		char := kindPart[ix]
+		if toUpper {
+			kind = kind + string([]byte{(char - 32)})
+			toUpper = false
+		} else if char == '-' {
+			toUpper = true
+		} else {
+			kind = kind + string([]byte{char})
+		}
+	}
+
+	group = strings.Join(parts[1:], ".")
+
+	return
+}
--- a/tpr/kind_test.go
+++ b/tpr/kind_test.go
@@ -9,35 +9,35 @@ import (
 
 func TestExtractKindAndGroup(t *testing.T) {
 	tests := []struct {
-		Input         string
-		ExpectedKind  string
-		ExpectedGroup string
-		ExpectedError error
+		name          string
+		expectedKind  string
+		expectedGroup string
+		expectedError error
 	}{
 		{
-			Input:         "foo.company.com",
-			ExpectedKind:  "Foo",
-			ExpectedGroup: "company.com",
+			name:          "foo.company.com",
+			expectedKind:  "Foo",
+			expectedGroup: "company.com",
 		},
 		{
-			Input:         "cron-tab.company.com",
-			ExpectedKind:  "CronTab",
-			ExpectedGroup: "company.com",
+			name:          "cron-tab.company.com",
+			expectedKind:  "CronTab",
+			expectedGroup: "company.com",
 		},
 		{
-			Input:         "foo",
-			ExpectedError: unexpectedlyShortResourceNameError,
+			name:          "foo",
+			expectedError: unexpectedlyShortResourceNameError,
 		},
 		{
-			Input:         "foo.company",
-			ExpectedError: unexpectedlyShortResourceNameError,
+			name:          "foo.company",
+			expectedError: unexpectedlyShortResourceNameError,
 		},
 	}
 
-	for i, tt := range tests {
-		kind, group, err := extractKindAndGroup(tt.Input)
-		assert.Equal(t, tt.ExpectedError, errgo.Cause(err), "#%d", i)
-		assert.Equal(t, tt.ExpectedKind, kind, "#%d", i)
-		assert.Equal(t, tt.ExpectedGroup, group, "#%d", i)
+	for i, tc := range tests {
+		kind, group, err := extractKindAndGroup(tc.name)
+		assert.Equal(t, tc.expectedError, errgo.Cause(err), "#%d", i)
+		assert.Equal(t, tc.expectedKind, kind, "#%d", i)
+		assert.Equal(t, tc.expectedGroup, group, "#%d", i)
 	}
 }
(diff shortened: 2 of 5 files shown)


tpr/kind.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+/*
+	Code based on:
+	https://github.com/kubernetes/kubernetes/blob/6a4d5cd7cc58e28c20ca133dab7b0e9e56192fe3/pkg/registry/extensions/thirdpartyresourcedata/util.go
+*/
+
+// extractKindAndGroup extracts kind and group from a name. For details see
+// Config.Name godoc.
+func extractKindAndGroup(name string) (kind string, group string, err error) {
```
asymmetric: What do you think about doing `(kind, group string, err error)`?

/bikeshedding

tpr/kind.go
```diff
@@ -0,0 +1,40 @@
...
+
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+/*
+	Code based on:
+	https://github.com/kubernetes/kubernetes/blob/6a4d5cd7cc58e28c20ca133dab7b0e9e56192fe3/pkg/registry/extensions/thirdpartyresourcedata/util.go
+*/
+
+// extractKindAndGroup extracts kind and group from a name. For details see
```
asymmetric: I think `For details, see **the**` would be better here.

tpr/kind.go
```diff
@@ -0,0 +1,40 @@
...
+// Config.Name godoc.
+func extractKindAndGroup(name string) (kind string, group string, err error) {
+	parts := strings.Split(name, ".")
+	if len(parts) < 3 {
+		return "", "", microerror.MaskAnyf(unexpectedlyShortResourceNameError, "%s, expected at least <kind>.<domain>.<tld>", name)
```
asymmetric: We've been using a different variable name in [aws-operator](https://github.com/giantswarm/aws-operator/blob/320bda04e31d246a172d6af81fb6656dd93807b5/service/create/load_balancer.go#L108), which I find reads a little bit better, and is more generic.

tpr/kind.go
```diff
@@ -0,0 +1,40 @@
...
+/*
+	Code based on:
+	https://github.com/kubernetes/kubernetes/blob/6a4d5cd7cc58e28c20ca133dab7b0e9e56192fe3/pkg/registry/extensions/thirdpartyresourcedata/util.go
+*/
+
+// extractKindAndGroup extracts kind and group from a name. For details see
+// Config.Name godoc.
+func extractKindAndGroup(name string) (kind string, group string, err error) {
+	parts := strings.Split(name, ".")
+	if len(parts) < 3 {
```
asymmetric: Would be cool to provide this as a constant.

tpr/kind.go
```diff
@@ -0,0 +1,40 @@
...
+		return "", "", microerror.MaskAnyf(unexpectedlyShortResourceNameError, "%s, expected at least <kind>.<domain>.<tld>", name)
+	}
+
+	// kind
+	kindPart := parts[0]
+	toUpper := true
+	for ix := range kindPart {
```
asymmetric: What does `ix` stand for?

tpr/kind.go
kopiczko: I'm ok with both.

tpr/kind_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"testing"
+
+	"github.com/juju/errgo"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExtractKindAndGroup(t *testing.T) {
+	tests := []struct {
+		Input         string
```
asymmetric: These don't really need to be exported

tpr/tpr.go
```diff
@@ -26,17 +26,15 @@ type Config struct {
 
 	// Settings.
 
-	// Name of the kind of ThirdPartyObjects. It should be in lower case
-	// and hyphen delimited. Kind names will be converted to CamelCase
-	// when creating ThirdPartyObjects. Hyphens in the kind are assumed to
-	// be word breaks. For instance the kind camel-case would be converted
-	// to CamelCase but camelcase would be converted to Camelcase.
+	// Name takes the form <kind>.<domain>. The domain is also known as
```
asymmetric: Can we settle on one of domain and group? What's the benefit of mentioning both?

tpr/tpr.go
```diff
@@ -51,13 +49,14 @@ type Config struct {
...
-	qualifiedName string // name.group
+	name        string // see Config.Name
+	kind        string // see Config.Name
+	group       string // see Config.Name
+	version     string
+	apiVersion  string // apiVersion is group/version
+	description string
 
-	endpointList string
+	endpoint string
```
asymmetric: Why does this change?

tpr/kind.go
kopiczko: I don't know:
https://github.com/kubernetes/kubernetes/blob/6a4d5cd7cc58e28c20ca133dab7b0e9e56192fe3/pkg/registry/extensions/thirdpartyresourcedata/util.go#L48

index perhaps? I can change it to `i`

tpr/tpr.go
```diff
@@ -51,13 +49,14 @@ type Config struct {
 type TPR struct {
 	clientset kubernetes.Interface
 
-	name          string
-	group         string
-	version       string
-	description   string
-	qualifiedName string // name.group
+	name        string // see Config.Name
```
asymmetric: What's the policy with adding comments after struct fields vs above? :police_car:

tpr/kind_test.go
asymmetric: Think this should be called `name`, to reflect the actual argument the function takes.

tpr/kind.go
```diff
@@ -0,0 +1,40 @@
...
+	if len(parts) < 3 {
+		return "", "", microerror.MaskAnyf(unexpectedlyShortResourceNameError, "%s, expected at least <kind>.<domain>.<tld>", name)
+	}
+
+	// kind
+	kindPart := parts[0]
+	toUpper := true
+	for i := range kindPart {
+		char := kindPart[i]
+		if toUpper {
```
asymmetric: I have no idea what is going on in this loop :)

Maybe some documentation would help.

tpr/kind.go
```diff
@@ -0,0 +1,40 @@
...
+			kind = kind + string([]byte{(char - 32)})
```
asymmetric: Remove the magic constant?

tpr/kind_test.go
```diff
@@ -0,0 +1,43 @@
...
+			Input:         "foo",
+			ExpectedError: unexpectedlyShortResourceNameError,
+		},
+		{
+			Input:         "foo.company",
+			ExpectedError: unexpectedlyShortResourceNameError,
+		},
+	}
+
+	for i, tt := range tests {
```
asymmetric: Elsewhere I've seen this variable be called `tc`, for test case. It would be nice to standardize here, whatever we end up choosing. I think go tests introduce enough boilerplate already, and standardizing reduces the overhead of separating that from the actual logic, IMO.

tpr/tpr.go
asymmetric: Also, is the `name` field needed anymore, or is it superseded by `kind` and `group`?

tpr/tpr.go
kopiczko: Because this endpoint can be used to more operations than only listing resource, this will be changed in my next PR anyway.

tpr/tpr.go
kopiczko: IDK

tpr/tpr.go
kopiczko: https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-third-party-resource/

^ it's so confusing here, I don't know which one to use to make it less confusing. Most of the comment is copied from the URL above.

Then in the same doc you see:

```
$ kubectl get crontab
NAME                 KIND
my-new-cron-object   CronTab.v1.stable.example.com
```

Then you think, what is name, kind, group, and domain?

It's obvious that group==domain. I don't know which one to use. I also think that when it's domain `cluster.giantswarm.io` doesn't make sense. 

You see, it's complicated :)

- - -

But ok, let's stick with group, bc I think this is what we use most of the time.
My proposition is to change godoc to 

```
Name takes the form <kind>.<group>. The group is also known as a domain. [...]
```

And no more `domain` word in this file.

WDYT?

kopiczko:
@rossf7 I added more description on top.

tpr/kind.go
kopiczko: 1. There's a test.
2. See Config.Name godoc. It's explained there how kind name is created from name.

tpr/kind.go
kopiczko: I just copied most of that code from https://github.com/kubernetes/kubernetes/blob/6a4d5cd7cc58e28c20ca133dab7b0e9e56192fe3/pkg/registry/extensions/thirdpartyresourcedata/util.go as the comment above says. `- 32` converts letter to lower case.

I have no strong opinion, let me know if blocker.

tpr/kind_test.go
kopiczko: I can change I used `tt` so much that I don't even use my brain to type that :)

tpr/tpr.go
kopiczko: Name is still used to create `v1.ObjectMeta` and annotate errors.

tpr/kind.go
kopiczko: Agree `malformedResourceNameError` is beautiful.

tpr/tpr.go
asymmetric: I would put that in `(Note: The group is sometimes also known as a domain)`

asymmetric: But yeah, otherwise I think it makes sense to stick to group.

tpr/kind.go
kopiczko: Can we leave it like that? IDK how to name it, and it makes sense for me to leave it, because it's immediately explained in error annotation below.

tpr/kind.go
rossf7: I'm OK with this for now as we're following the k8s code. We can improve it over time plus TPR support is still alpha in k8s.

tpr/tpr.go
rossf7: IDK either so this would be good to have in the Go style guide.

tpr/tpr.go
```diff
@@ -93,11 +92,13 @@ func New(config Config) (*TPR, error) {
 	return tpr, nil
 }
 
-// Kind returns TPR kind extracted from Name. See Config.Name godoc for details.
+// Kind returns TPR kind extracted from Name. Useful when creating
```
asymmetric: You need an article after "returns" IMO :) Same above and below.

kopiczko: I'm bad at this :)

asymmetric: It's normal I guess for slavic people right? :)

kopiczko: I don't know, but interesting :)

PR #16, opened 2017-05-26
Endpoint

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/1483

Adds Endpoint and WatchEndpoint methods to TPR object

### DIFF
--- a/tpr/kind.go
+++ b/tpr/kind.go
@@ -40,7 +40,7 @@ func extractKindAndGroup(name string) (kind, group string, err error) {
 }
 
 // unpluralizedSuffixes is a list of resource suffixes that are the same plural
-// and singular. This is only necessary because some bits of [kubernetes] code
+// and singular. This is only necessary because some bits of (kubernetes) code
 // are lazy and don't actually use the RESTMapper like they should.
 //
 // Copied from:
--- a/tpr/kind.go
+++ b/tpr/kind.go
@@ -39,9 +39,10 @@ func extractKindAndGroup(name string) (kind, group string, err error) {
 	return
 }
 
-// unpluralizedSuffixes is a list of resource suffixes that are the same plural
-// and singular. This is only necessary because some bits of (kubernetes) code
-// are lazy and don't actually use the RESTMapper like they should.
+// unpluralizedSuffixes is a list of resource suffixes that are the same when
+// plural and singular. This is only necessary because some bits of
+// (kubernetes) code are lazy and don't actually use the RESTMapper like they
+// should.
 //
 // Copied from:
 // https://github.com/kubernetes/kubernetes/blob/b0b711119b48854e0b73805e42be2bcc4b2bd604/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L131-L137
--- a/tpr/kind.go
+++ b/tpr/kind.go
@@ -38,3 +38,39 @@ func extractKindAndGroup(name string) (kind, group string, err error) {
 
 	return
 }
+
+// unpluralizedSuffixes is a list of resource suffixes that are the same plural
+// and singular This is only is only necessary because some bits of code are
+// lazy and don't actually use the RESTMapper like they should.
+//
+// Copied from:
+// https://github.com/kubernetes/kubernetes/blob/b0b711119b48854e0b73805e42be2bcc4b2bd604/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L131-L137
+var unpluralizedSuffixes = []string{
+	"endpoints",
+}
+
+// unsafeGuessKindToResource converts Kind to a plural resource name. It is
+// named after and has similar semantics to:
+// https://github.com/kubernetes/kubernetes/blob/b0b711119b48854e0b73805e42be2bcc4b2bd604/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L139-L164
+func unsafeGuessKindToResource(kind string) string {
+	if len(kind) == 0 {
+		return ""
+	}
+
+	resource := strings.ToLower(kind)
+
+	for _, skip := range unpluralizedSuffixes {
+		if strings.HasSuffix(resource, skip) {
+			return resource
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


tpr/kind.go
```diff
@@ -38,3 +38,39 @@ func extractKindAndGroup(name string) (kind, group string, err error) {
 
 	return
 }
+
+// unpluralizedSuffixes is a list of resource suffixes that are the same plural
+// and singular This is only is only necessary because some bits of code are
```
asymmetric: Repetition typo

tpr/tpr.go
```diff
@@ -55,7 +54,9 @@ type TPR struct {
 	apiVersion  string // apiVersion is group/version
 	description string
 
-	endpoint string
+	// resourceName is name of a resource registered in Kubernetes API for
```
asymmetric: is the name

tpr/tpr.go
```diff
@@ -107,6 +108,28 @@ func (t *TPR) Name() string { return t.name }
 // Group returns a TPR group extracted from Name. See Config.Name godoc for details.
 func (t *TPR) Group() string { return t.group }
 
+// Endpoint returns a TPR resurce endpoint registered in Kubernetes API under
```
asymmetric: in the K8s API

tpr/tpr.go
```diff
@@ -107,6 +108,28 @@ func (t *TPR) Name() string { return t.name }
...
+// a given namespace. Default namespace will be used when argument is an empty
```
asymmetric: The default namespace ... when the argument ;)

tpr/tpr.go
```diff
@@ -107,6 +108,28 @@ func (t *TPR) Name() string { return t.name }
...
+// string.
+func (t *TPR) Endpoint(namespace string) string {
+	nsResource := t.resourceName
+	if len(namespace) != 0 {
+		nsResource = "namespace/" + namespace + "/" + t.resourceName
+	}
+	return "/apis/" + t.group + "/" + t.version + "/" + nsResource
```
asymmetric: Can we use `strings.Join` here instead?

kopiczko: I thought about path.Join already. Problem is that we'd have to create different slice depending if namespace is `""` or not.

kopiczko: There are also tests to make sure there aren't missing or redundant slashes.

tpr/tpr.go
```diff
@@ -119,9 +119,9 @@ func (t *TPR) Endpoint(namespace string) string {
 	return "/apis/" + t.group + "/" + t.version + "/" + nsResource
 }
 
-// Endpoint returns a TPR watch resurce endpoint registered in Kubernetes API
-// under a given namespace. Default namespace will be used when argument is an
-// empty string.
+// Endpoint returns a TPR watch resurce endpoint registered in the Kubernetes
```
asymmetric: resource

tpr/tpr.go
```diff
@@ -119,9 +119,9 @@ func (t *TPR) Endpoint(namespace string) string {
...
+// API under a given namespace. Default namespace will be used when the
```
asymmetric: the default (above too)

tpr/kind.go
```diff
@@ -40,8 +40,8 @@ func extractKindAndGroup(name string) (kind, group string, err error) {
 }
 
 // unpluralizedSuffixes is a list of resource suffixes that are the same plural
-// and singular This is only is only necessary because some bits of code are
-// lazy and don't actually use the RESTMapper like they should.
+// and singular. This is only necessary because some bits of [kubernetes] code
```
asymmetric: Are the `[]` there for a link? Otherwise I'd convert to `()`.

kopiczko:
@asymmetric 

> Might be worth extracting this pluralization stuff to its own library at some point.

Maybe. It's shitty at the moment :P And it must be like that to match K8s API. When TPR will be updated to [allow specifying custom plural form](https://github.com/giantswarm/giantswarm/issues/1483#issuecomment-303140694), that may make more sense as people maybe will be adding exceptions.

tpr/kind.go
```diff
@@ -38,3 +38,39 @@ func extractKindAndGroup(name string) (kind, group string, err error) {
 
 	return
 }
+
+// unpluralizedSuffixes is a list of resource suffixes that are the same plural
```
rossf7: Nitpick but I think "when plural and singular" is easier to read.

tpr/kind_test.go
```diff
@@ -41,3 +41,32 @@ func TestExtractKindAndGroup(t *testing.T) {
 		assert.Equal(t, tc.expectedGroup, group, "#%d", i)
 	}
 }
+
+func TestUnsafeGuessKindToResource(t *testing.T) {
```
rossf7: <3 that the tests cover all the exceptions.

tpr/tpr.go
```diff
@@ -55,7 +54,9 @@ type TPR struct {
 	apiVersion  string // apiVersion is group/version
 	description string
 
-	endpoint string
+	// resourceName is the name of a resource registered in Kubernetes API
```
rossf7: "in the Kubernetes API"

tpr/tpr.go
```diff
@@ -107,6 +108,28 @@ func (t *TPR) Name() string { return t.name }
...
+func (t *TPR) Endpoint(namespace string) string {
+	nsResource := t.resourceName
+	if len(namespace) != 0 {
+		nsResource = "namespace/" + namespace + "/" + t.resourceName
+	}
+	return "/apis/" + t.group + "/" + t.version + "/" + nsResource
+}
+
+// Endpoint returns a TPR watch resource endpoint registered in the Kubernetes
+// API under a given namespace. Default namespace will be used when the
```
rossf7: "The default namespace"

kopiczko:
@rossf7 yes, but this is a library, it should also support 1.6 for a while.

kopiczko:
Agh..., sorry for the commit message. Clicked before read..

PR #28, opened 2017-06-28
fix setting resyncPeriod in TPR

kopiczko:


tpr/tpr.go
```diff
@@ -109,12 +109,13 @@ func New(config Config) (*TPR, error) {
...
-		version:     config.Version,
-		apiVersion:  group + "/" + config.Version,
-		description: config.Description,
+		name:         config.Name,
+		kind:         kind,
+		group:        group,
+		version:      config.Version,
+		apiVersion:   group + "/" + config.Version,
+		description:  config.Description,
+		resyncPeriod: config.ResyncPeriod,
```
kopiczko: This is the only line I added, rest is the indentation.

PR #33, opened 2017-07-13
update client-go to v3.0.0

kopiczko:
* update to stable version
* pin apimachinery to the version supported by client-go


glide.yaml
```diff
@@ -3,12 +3,17 @@ import:
...
 - package: github.com/giantswarm/microkit
+  version: master
   subpackages:
   - error
   - logger
 - package: github.com/juju/errgo
+  version: master
 - package: k8s.io/client-go
-  version: v3.0.0-beta.0
+  version: v3.0.0
```
kopiczko: `-beta.0` is dropped. There is also `v4.0.0-beta.0`, but I don't want to deal with it ATM. Probably introduces CRD support.

glide.yaml
```diff
@@ -3,12 +3,17 @@ import:
...
+# apimachinery is pinned to the version client-go uses.
+- package: k8s.io/apimachinery
+  version: 85ace5365f33b16fc735c866a12e3c765b9700f2
```
kopiczko: I also pinned apimachinery. `master` doesn't work with `v3.0.0`. It's better do to it here once than in every project using operatorkit.

PR #56, opened 2017-09-04
add aws client

fgimenez:
Towards giantswarm/aws-operator/issues/381

Migrated from aws-operator


fgimenez:
@kopiczko of course, the service will be moved to microendpoint, this is an initial step, aws-operator is currently using github.com/giantswarm/aws-operator/client/aws for checking the health https://github.com/giantswarm/aws-operator/blob/master/service/healthz/service.go#L78, if we just move the service to microendpoint it would end depending on aws-operator, which doesn't make too much sense to me (please correct me if I'm wrong). That's why this PR proposes to move that client to a repo from which it can be imported from both microendpoint (for the aws health check) and aws-operator to keep using it. 

The k8s client in operatorkit is an example of the same pattern (it pulls the whole k8s.io/client-go dependency tree), let me know WDYT about this approach.

kopiczko:
> The k8s client in operatorkit is an example of the same pattern

You're right, but we can't make operatorkit an aggregator of random API clients. It's meant to make creating operators easier. We don't want to make every operator depend on AWS.

I understand you followed the pattern, but I think it'd be better to add a client package in microendpoint. I'm also fine with aws-operator creating awsclient using microendpoint package, until we come with some better solution. I created an issue here: https://github.com/giantswarm/microendpoint/issues/11

kopiczko:
Or maybe even stop proceeding with this before https://github.com/giantswarm/microendpoint/issues/11 is solved. Do you have something else on your plate?

rossf7:
@kopiczko this was my suggestion to use `operatorkit` but you're right we should keep it lean. Operatorkit should be k8s focused and we'd like the community to adopt it.

Having the AWS client in `microendpoint` could be an interim step but there is still the deps issue. Maybe we do need separate repos for these healthchecks?

kopiczko:
@rossf7 

> Maybe we do need separate repos for these healthchecks

Yeah, I'd be fine with that. That's what I described in https://github.com/giantswarm/microendpoint/issues/11

> Having the AWS client in microendpoint could be an interim step but there is still the deps issue

Yes, but it'd be just moving the code around. If we create a separate repo, then the code'd move again. I'm afraid it'd be a waste of time.

fgimenez:
> Or maybe even stop proceeding with this before giantswarm/microendpoint#11 is solved. Do you have something else on your plate?

@kopiczko I can go ahead with it, as it blocks the issue I'm working on, closing this one.

@kopiczko @rossf7 thanks! will ping you when ready, any suggestion for the repos' names? also, should  giantswarm/microendpoint#11 be added to the board?

kopiczko:
@fgimenez I already added it to sig operator. Re naming: for `microstorage` we have `tprstorage` and `etcdstorage`. Maybe that inspires you :D

PR #57, opened 2017-09-05
client: make K8s client in-cluster by default

kopiczko:


client/k8s/k8s.go
```diff
@@ -54,8 +54,8 @@ func DefaultConfig() Config {
 		Logger: newLogger,
 
 		// Settings.
-		Address:   "http://127.0.0.1:8080",
-		InCluster: false,
+		Address:   "",
```
fgimenez: Given that "" is the zero value for strings, could we remove this?

kopiczko: Talk to @xh3b4sd when he's back. He likes that.

PR #94, opened 2017-10-25
Add retry on creating the tpr as well

oponder:
This will make test environments where the kubernetes-api is started up along with services that want to talk to it right away more reliable.

### DIFF
--- a/tpr/tpr.go
+++ b/tpr/tpr.go
@@ -202,7 +202,7 @@ func (t *TPR) CreateAndWait() error {
 // poll for TPR readiness. Returns alreadyExistsError when the resource already
 // exists.
 func (t *TPR) CreateAndWaitBackOff(initBackOff backoff.BackOff) error {
-	err := t.create()
+	err := t.create(initBackOff)
 	if err != nil {
 		return microerror.Maskf(err, "creating TPR %s", t.name)
 	}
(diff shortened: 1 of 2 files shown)


oponder:
There is tiny bit more complexity here than simply wrapping the create call and using backoff on that.

Since I wan't to 'cancel' the backoff if the create error is due to the TPR already existing.

In some way this reveals that the 'already exists' error is not really an error from the point of view of the create function. Not sure if my approach is the best way to solve this.

My approach was to first try creating the TPR without doing any retries. If that fails because of the TPR already existing, then I do not continue to the code that attempts to create the TPR with _retries_.

tpr/tpr.go
```diff
@@ -263,13 +263,36 @@ func (t *TPR) create() error {
...
+		}
+		if err != nil {
+			return microerror.Maskf(err, "creating TPR %s", t.name)
+		}
+
+		return nil
+	}
+
+	// Try creating the TPR once without a retry, since I
+	// do not want to retry on the 'alreadyExistsError'.
```
kopiczko: You have PermanentError for that https://github.com/cenkalti/backoff/blob/309aa717adbf351e92864cbedf9cca0b769a4b5a/retry.go#L74

kopiczko: I think you'll have to switch backoff version to master in glide.yaml. Which should be done anyway.

oponder:
Look at me contributing to operatorkit! pew pew. what a time to be alive!

PR #107, opened 2017-11-16
Adds tpr metrics

JosephSalisbury:


tpr/tpr.go
```diff
@@ -297,3 +298,35 @@ func (t *TPR) waitInit(retry backoff.BackOff) error {
 	}
 	return microerror.Maskf(err, "requesting TPR %s", t.name)
 }
+
+func (t *TPR) CollectMetrics() {
+	go func() {
+		t.logger.Log("info", "starting metrics collection")
+
+		ticker := time.NewTicker(t.resyncPeriod)
+		for range ticker.C {
```
fgimenez: It would be super nice to have the chance to not leak this goroutine, maybe letting `CollectMetrics` accept a context and having something like:
```
for {
  select {
    case <-ctx.Done():
      return ""
    case <-ticker.C:
      ...
  }
}
```
The function where `f.tpr.CollectMetrics()` is called injects contexts in other calls, they are `TODOs` though, not sure if they are being used, but having it in place would be good imo.

JosephSalisbury:
@fgimenez mind having another look?

PR #132, opened 2017-12-08
informer: simplify watcher interface

kopiczko:


informer/informer_integration_test.go
```diff
@@ -136,59 +133,22 @@ func init() {
...
+	if err != nil {
+		panic(fmt.Sprintf("%#v", err))
 	}
 }
 
 func testNewInformer(t *testing.T, rateWait, resyncPeriod time.Duration) *Informer {
 	c := DefaultConfig()
 
-	c.WatcherFactory = newWatcherFactory
+	c.Watcher = k8sClient.CoreV1().ConfigMaps(namespace)
```
kopiczko: This feels so goooood.

informer/informer_integration_test.go
```diff
@@ -20,30 +20,30 @@ package informer
 */
```
kopiczko: The test watches for ConfigMaps in a namespace. So it don't have to ensure CRD and cleanup is simply namespace deletion.

informer/decoder.go
```diff
@@ -1,41 +0,0 @@
-package informer
```
kopiczko: No decoder :tada:

informer/zero_object.go
```diff
@@ -1,19 +0,0 @@
-package informer
```
kopiczko: No zero_object :tada:

PR #133, opened 2017-12-09
dep: replace glide

kopiczko:
This is a copy of https://github.com/giantswarm/operatorkit/pull/131 which by mistake I merged to wrong branch.


kopiczko:
Going with that as this is already reviewed.

PR #137, opened 2017-12-12
dep: update kubernetes deps to 1.8.5

kopiczko:


kopiczko:
Going with this as this is a vendor update.

PR #154, opened 2018-02-09
framework: put CRUDResourceOps in separate file

kopiczko:


framework/crud_resource_ops.go
```diff
@@ -0,0 +1,90 @@
+package framework
+
+import "context"
+
```
kopiczko: 100% copy paste.

PR #162, opened 2018-02-27
framework: improve router error messages

kopiczko:


framework/resource_router.go
```diff
@@ -37,10 +37,10 @@ func (r *ResourceRouter) ResourceSet(obj interface{}) (*ResourceSet, error) {
 	}
 
 	if len(found) == 0 {
-		return nil, microerror.Maskf(executionFailedError, "could not find any resource router")
+		return nil, microerror.Maskf(executionFailedError, "could not find handling resource set")
```
tuommaki: Nitpick, but for me this somehow sounds weird. What would think of following:
`no match for resource set`

framework/resource_router.go
```diff
@@ -37,10 +37,10 @@ func (r *ResourceRouter) ResourceSet(obj interface{}) (*ResourceSet, error) {
...
 	}
 	if len(found) > 1 {
-		return nil, microerror.Maskf(executionFailedError, "multiple resource routers not allowed")
+		return nil, microerror.Maskf(executionFailedError, "multiple handling resource sets found")
```
tuommaki: Along the lines of above, my suggestion for this:
`multiple matches for resource set; only single allowed`

framework/resource_router.go
kopiczko: This means that resource set that is capable of handling custom object that just arrived wasn't found.

I guess is sounds more reasonable when you are into the details, doesn't it?

I'd be glad to make it something reasonable for a newcomer. `handling resource set not found` makes more sense?

tuommaki: Yeah, I actually did know what this is used for. Your new suggestion sounds very good. I like it :+1:

framework/resource_router.go
kopiczko: Changing to `multiple handling resource sets found; only single allowed` to align with `handling resource set not found`.

PR #177, opened 2018-03-27
Align circle config

MarcelMue:

Should be pretty straight forward, integration tests seem to work in parallel now.


framework/finalizer.go
```diff
@@ -50,6 +50,45 @@ func (f *Framework) addFinalizer(obj interface{}) (stopReconciliation bool, err
 	return true, nil
 }
 
+func (f *Framework) removeFinalizer(ctx context.Context, obj interface{}) error {
```
MarcelMue: Just some quick boyscouting to have methods before functions, nothing changed.

PR #189, opened 2018-04-12
Fix flakyness in reconciliation integration testcase

MarcelMue:
This is a simple change that caused flakyness in the test that will need a little bit of explaining:

The framework is configured to reconcile every 10 seconds. This tests counts how often the resource functions of the framework have been called and checks if that's correct.

This test always waits a set amount of time before the check, this was 25 seconds. So 5 seconds to do the operation (add or delete an object) with a lot of room for safety and then two reconciliation periods.

So the overall time it ran was 2*25 + test execution time = 50 to 51 seconds. This lead to the problem: if execution was a little slower but the framework executed very quickly, then it could reconcile 5 times in roughly 51 seconds the test ran. This would cause the test to fail.

So I reduced the wait each time by 2 seconds, so we wait 2*23 = 46 seconds in total now. So we still have a 6 second buffer in total, but are also safe with 4 seconds before the framework reconciles a fifth time.

UPDATE: Using backoff now with 24 seconds, 23 seconds was too little atleast once for me. Really hoping that this is more stable.

TL;DR:
Reducing the safety time individually makes the whole test more stable.

### DIFF
--- a/framework/integration/test/reconciliation/reconciliation_test.go
+++ b/framework/integration/test/reconciliation/reconciliation_test.go
@@ -92,7 +92,7 @@ func Test_Finalizer_Integration_Reconciliation(t *testing.T) {
 	}
 
 	// We wait the absolute maximum amount of time here:
-	// 20 second ResyncPeriod + 2 second RateWait + 3 second for safety.
+	// 20 second ResyncPeriod + 2 second RateWait + 1 second for safety.
 	// The framework should now add the finalizer and EnsureCreated should be hit
 	// once immediatly.
 	//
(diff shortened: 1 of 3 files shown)


framework/integration/test/reconciliation/reconciliation_test.go
```diff
@@ -102,7 +102,7 @@ func Test_Finalizer_Integration_Reconciliation(t *testing.T) {
 	//
 	// 		EnsureCreated: 3, EnsureDeleted: 0
 	//
-	time.Sleep(25 * time.Second)
+	time.Sleep(23 * time.Second)
```
kopiczko: I think it can be improved by using backoff and checking CreateCount there. After that you can check it was ~22 seconds since test start and all other stuff.

MarcelMue: You are right, that's probably better. Going to do this.

MarcelMue: I changed this, but it still looks a bit ugly. I will think of something to structure this a bit better.

MarcelMue:
PTAL @kopiczko

PR #192, opened 2018-04-16
framework: always set reconciliationcanceledcontext

kopiczko:


framework/resource_set.go
```diff
@@ -67,7 +66,6 @@ func NewResourceSet(c ResourceSetConfig) (*ResourceSet, error) {
 }
 
 func (r *ResourceSet) InitCtx(ctx context.Context, obj interface{}) (context.Context, error) {
-	ctx = reconciliationcanceledcontext.NewContext(ctx, make(chan struct{}))
 	ctx = updateallowedcontext.NewContext(ctx, make(chan struct{}))
 	ctx = updatenecessarycontext.NewContext(ctx, make(chan struct{}))
```
kopiczko: Now resource-related context are set in resource_set.go, and reconciliation-related contexts are set in framework.go.

framework/framework.go
```diff
@@ -339,6 +341,8 @@ func ProcessUpdate(ctx context.Context, obj interface{}, resources []Resource) e
 		return microerror.Maskf(executionFailedError, "resources must not be empty")
 	}
 
+	ctx = reconciliationcanceledcontext.NewContext(ctx, make(chan struct{}))
+
```
kopiczko: Moved here from resource_set.go

framework/framework_cancel_crud_test.go
```diff
@@ -102,59 +97,9 @@ func Test_ProcessDelete_CRUD(t *testing.T) {
...
-
-				"r4.GetCurrentState",
-				"r4.GetDesiredState",
-				"r4.NewDeletePatch",
-				"r4.ApplyCreateChange",
-				"r4.ApplyDeleteChange",
-				"r4.ApplyUpdateChange",
-			},
-			ErrorMatcher: nil,
-		},
```
kopiczko: There is no such case now, which I believe was confusing and inconsistent with resource cancellation.

framework/framework_cancel_test.go
```diff
@@ -113,7 +87,7 @@ func Test_ProcessDelete(t *testing.T) {
 	}
 
 	for i, tc := range testCases {
-		err := ProcessDelete(tc.Ctx, nil, tc.Resources)
+		err := ProcessDelete(context.Background(), nil, tc.Resources)
```
kopiczko: I tend to use `context.Background()` instead `context.TODO()` in tests, because it isn't really TODO and this messes with greping for TODOs.

PR #204, opened 2018-05-03
increase wait times in backoff for integration tests

MarcelMue:

### DIFF
--- a/controller/integration/test/controlflow/controlflow_test.go
+++ b/controller/integration/test/controlflow/controlflow_test.go
@@ -86,7 +86,7 @@ func Test_Finalizer_Integration_Controlflow(t *testing.T) {
 	}
 
 	// We use backoff with the absolute maximum amount:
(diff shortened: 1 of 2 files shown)


MarcelMue:
I reran integration tests here 3 times and nothing failed. Might be a fluke but hopefully this is an improvement.

MarcelMue:
No we use `backoff` with a one second linear check. So we always stop and move on to the next block in our test if we reach a specific count after at most 1 second. 

The controller might get an event right away and we wait for one more reconciliation period:
~1 + 10 = 11 seconds
It might not get the event right away but we still wait for the reconciliation:
~4 + 10 = 14 seconds

In both cases the operator _just_ reconciled, so we have 10 seconds until the next one and can move on in the test without getting too many counts.

MarcelMue:
Going with this, ping me if there are questions.

PR #205, opened 2018-05-03
Add multi framework integration test for finalizers

MarcelMue:

### DIFF
--- a/controller/integration/test/parallel/parallel_test.go
+++ b/controller/integration/test/parallel/parallel_test.go
@@ -31,7 +31,7 @@ func Test_Finalizer_Integration_Parallel(t *testing.T) {
 	testFinalizerB := "operatorkit.giantswarm.io/test-operator-b"
 	operatorNameB := "test-operator-b"
 
-	// We create the first controller "A" here with its own resource.
+	// We create the first resource "A" here with its own resource.
 	var trA *testresource.Resource
 	{
 		c := testresource.Config{}
@@ -49,7 +49,7 @@ func Test_Finalizer_Integration_Parallel(t *testing.T) {
 
 	controllerA := testWrapperA.Controller()
 
-	// We create the second controller "B" and give it a different resource.
+	// We create the second resource "B" and give it a different resource.
 	var trB *testresource.Resource
 	{
 		c := testresource.Config{}
--- a/controller/integration/test/parallel/parallel_test.go
+++ b/controller/integration/test/parallel/parallel_test.go
@@ -165,7 +165,7 @@ func Test_Finalizer_Integration_Parallel(t *testing.T) {
 	}
 
 	// We use backoff with the absolute maximum amount:
-	// 10 second ResyncPeriod + 2 second RateWait + 8 second for safety.
+	// 2 second RateWait + 18 second to ensure that both finalizers were removed.
 	// The controllers should now remove the finalizer and EnsureDeleted should be
 	// hit four times immediatly.
 	// See https://github.com/giantswarm/giantswarm/issues/2897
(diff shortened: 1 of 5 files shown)


controller/integration/test/parallel/parallel_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+	controllerB := testWrapperB.Controller()
+
+	// We setup the namespace in which we test. We use the wrapper of controller A
+	// here, it makes no difference if we use the wrapper of A or B.
+	testWrapperA.MustSetup(testNamespace)
+	defer testWrapperA.MustTeardown(testNamespace)
+
+	// We start the controllers.
+	go controllerA.Boot()
```
MarcelMue: Note that both controllers have their own informer in this setup. Would it be better to pass the same informer to both? 

I am unsure how much this changes or what we do more often in our operators.

controller/integration/test/parallel/parallel_test.go
```diff
@@ -0,0 +1,224 @@
...
+	// Creation is retried because the existance of a CRD might have to be ensured.
+	obj := &v1alpha1.NodeConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      objName,
+			Namespace: testNamespace,
+		},
+	}
+	var createdObj interface{}
+	operation := func() error {
+		createdObj, err = testWrapperA.CreateObject(testNamespace, obj)
```
MarcelMue: The testwrapper here uses the k8sclient or g8sclient internally, it doesn't matter if we use `A` or `B` as both are configured the same way. I stick with `A` throughout this test for consistency.

controller/integration/test/parallel/parallel_test.go
```diff
@@ -0,0 +1,224 @@
...
+	// We use backoff with the absolute maximum amount:
+	// 2 second RateWait + 18 second to ensure that both finalizers were removed.
+	// The controllers should now remove the finalizer and EnsureDeleted should be
+	// hit four times immediatly.
+	// See https://github.com/giantswarm/giantswarm/issues/2897
+	//
+	// 		EnsureDeleted: 4
+	//
+	operation = func() error {
+		if trA.DeleteCount() != 4 {
```
MarcelMue: We can check accurately here because there is no reconciliation.
We do not care about the `CreateCount` here anymore.

controller/integration/test/parallel/parallel_test.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		t.Fatal("expected", nil, "got", err)
+	}
+
+	// We verify that our object is completely gone now.
+	_, err = testWrapperA.GetObject(objName, testNamespace)
+	if !errors.IsNotFound(err) {
+		t.Fatalf("error == %#v, want NotFound error", err)
+	}
+}
```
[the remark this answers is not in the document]
MarcelMue: My idea here was to create a test with two actual controllers, so we can be sure that we have no race when we merge a new change.

The behaviour you describe is done in the `reconciliation` testcase where I add another finalizer and then test if the operator only removes its own finalizer. Or am I missing something?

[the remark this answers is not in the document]
xh3b4sd: Ok cool. All good.

controller/integration/test/parallel/parallel_test.go
```diff
@@ -0,0 +1,224 @@
...
+	controllerB := testWrapperB.Controller()
+
+	// We setup the namespace in which we test. We use the wrapper of controller A
+	// here, it makes no difference if we use the wrapper of A or B.
+	testWrapperA.MustSetup(testNamespace)
+	defer testWrapperA.MustTeardown(testNamespace)
+
+	// We start the controllers.
+	go controllerA.Boot()
+	go controllerB.Boot()
```
kopiczko: Why those are called as goroutines?

MarcelMue: The framework would block otherwise waiting for new events. With the goroutines I can continue the test and have the frameworks do their thing in parallel.

kopiczko: That sucks, but it is how it is now.

PR #207, opened 2018-05-08
Initial setup for the example operator

rossf7:
Pawel, here is a start on the example operator. I kept it small as I'd like to check the direction is OK.

### DIFF
--- a/example/memcached-operator/main.go
+++ b/example/memcached-operator/main.go
@@ -3,8 +3,10 @@ package main
 import (
 	"flag"
 	"fmt"
+	"os"
 
 	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
 
 	"github.com/giantswarm/operatorkit/example/memcached-operator/memcached"
 )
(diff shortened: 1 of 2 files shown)


example/memcached-operator/main.go
```diff
@@ -0,0 +1,56 @@
...
+	var newLogger micrologger.Logger
+	{
+		c := micrologger.Config{
+			IOWriter: os.Stdout,
+		}
+		newLogger, err = micrologger.New(c)
+		if err != nil {
+			return microerror.Maskf(err, "micrologger.New")
+		}
+	}
```
rossf7: I'm using micrologger here because it's used in OperatorKit. I know we want to have a text logger rather than JSON. Should we make that configurable in micrologger?

example/memcached-operator/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+func parseFlags() memcached.Config {
+	var config memcached.Config
+
+	flag.StringVar(&config.K8sAddress, "kubernetes.address", "", "Address used to connect to Kubernetes.")
+	flag.BoolVar(&config.K8sInCluster, "kubernetes.incluster", true, "Whether to use the in-cluster config to authenticate with Kubernetes.")
+	flag.StringVar(&config.K8sCAFile, "kubernetes.ca", "", "Certificate authority file path to use to authenticate with Kubernetes.")
+	flag.StringVar(&config.K8sCrtFile, "kubernetes.crt", "", "Certificate file path to use to authenticate with Kubernetes.")
+	flag.StringVar(&config.K8sKeyFile, "kubernetes.key", "", "Key file path to use to authenticate with Kubernetes.")
+	flag.Parse()
```
rossf7: I went with simpler flag handling based on the previous workshop rather than using Cobra.

PR #260, opened 2018-12-21
controller: fix multifinalizer test

kopiczko:
Code fixing test introduced in https://github.com/giantswarm/operatorkit/pull/259.

PR #265, opened 2019-01-02
controlflow: fix potential races

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/4572.

Pulled parts of the changes from https://github.com/giantswarm/operatorkit/pull/260.

### DIFF
--- a/controller/integration/test/controlflow/controlflow_test.go
+++ b/controller/integration/test/controlflow/controlflow_test.go
@@ -86,7 +86,7 @@ func Test_Finalizer_Integration_Controlflow(t *testing.T) {
 
 			return nil
 		}
-		b := backoff.NewExponential(2*time.Second, 10*time.Second)
+		b := backoff.NewExponential(10*time.Second, 1*time.Second)
 
 		err = backoff.Retry(o, b)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


controller/integration/test/controlflow/controlflow_test.go
```diff
@@ -83,158 +83,155 @@ func Test_Finalizer_Integration_Controlflow(t *testing.T) {
...
 	//
-	// 		EnsureCreated: 1, EnsureDeleted: 0
+	// 		EnsureCreated: >2, EnsureDeleted: =0
 	//
-	// The controller should reconcile once in this period.
-	//
-	// 		EnsureCreated: 2, EnsureDeleted: 0
-	//
-	operation := func() error {
-		if tr.CreateCount() != 2 {
```
kopiczko: We can't really rely on equality on those tests other than making sure that something is zero (i.e. situation we expect thing to not happen at all).

controller/integration/testresource/testresource.go
```diff
@@ -81,6 +89,8 @@ func (r *Resource) Name() string {
 }
 
 func (r *Resource) SetReturnErrorFunc(f func(obj interface{}) error) {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
 	r.returnErrorFunc = f
```
kopiczko: Maybe chances are low but it can be changed during actual reconciliation.

PR #280, opened 2019-01-24
controller: add more logs around event processing

kopiczko:
@xh3b4sd noticed there are missing loops, e.g. we have logs with
"loop":"25" and "loop":"27" but no logs with "loop":"26". In some rare
cases we see that controller stops reconciling. We need more diagnostic
information.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -300,8 +300,6 @@ func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Ev
 		case <-ctx.Done():
 			return nil
 		}
-
-		c.logger.LogCtx(ctx, "level", "debug", "message", "reconciled event")
 	}
 }
 
@@ -417,7 +415,7 @@ func (c *Controller) bootWithError(ctx context.Context) error {
 			return microerror.Mask(err)
 		}
 
-		c.logger.LogCtx(ctx, "level", "debug", "message", "stopped processing object events")
+		c.logger.LogCtx(ctx, "level", "debug", "message", "processed processing object events")
 	}
 
 	return nil
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -232,9 +232,6 @@ func (c *Controller) deleteFunc(ctx context.Context, obj interface{}) {
 // ProcessEvents takes the event channels created by the operatorkit informer
 // and executes the controller's event functions accordingly.
 func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Event, updateChan chan watch.Event, errChan chan error) error {
-	c.logger.LogCtx(ctx, "level", "debug", "message", "processing object events")
-	defer c.logger.LogCtx(ctx, "level", "debug", "message", "stopped processing object events")
-
 	loop := -1
 
 	for {
@@ -412,10 +409,15 @@ func (c *Controller) bootWithError(ctx context.Context) error {
 
 		deleteChan, updateChan, errChan := c.informer.Watch(ctx)
 		close(c.booted)
+
+		c.logger.LogCtx(ctx, "level", "debug", "message", "processing object events")
+
 		err := c.ProcessEvents(ctx, deleteChan, updateChan, errChan)
 		if err != nil {
 			return microerror.Mask(err)
 		}
+
+		c.logger.LogCtx(ctx, "level", "debug", "message", "stopped processing object events")
 	}
 
 	return nil
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -234,6 +234,9 @@ func (c *Controller) deleteFunc(ctx context.Context, obj interface{}) {
 func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Event, updateChan chan watch.Event, errChan chan error) error {
 	loop := -1
 
+	c.logger.LogCtx(ctx, "level", "debug", "message", "processing object events")
+	defer c.logger.LogCtx(ctx, "level", "debug", "message", "stopped processing object events")
+
 	for {
 		loop++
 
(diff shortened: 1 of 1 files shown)


controller/controller.go
```diff
@@ -232,6 +232,9 @@ func (c *Controller) deleteFunc(ctx context.Context, obj interface{}) {
 // ProcessEvents takes the event channels created by the operatorkit informer
 // and executes the controller's event functions accordingly.
 func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Event, updateChan chan watch.Event, errChan chan error) error {
+	c.logger.LogCtx(ctx, "level", "debug", "message", "processing object events")
```
xh3b4sd: The odd thing now is that this log will not contain the loop, wile the last one probably will. I would add this log down right after the loop got added to the context and keep the deferred log as it is. Then the result you observe at operator runtime will be consistent and helpful.

controller/controller.go
```diff
@@ -275,14 +280,16 @@ func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Ev
...
+					c.logger.LogCtx(ctx, "level", "error", "message", fmt.Sprintf("failed to create accessor %T", e.Object), "stack", fmt.Sprintf("%#v", err))
 				} else {
 					ctx = setLoggerCtxValue(ctx, loggerKeyObject, accessor.GetSelfLink())
 					ctx = setLoggerCtxValue(ctx, loggerKeyVersion, accessor.GetResourceVersion())
 				}
 			}
 
+			c.logger.LogCtx(ctx, "level", "debug", "message", "reconciling object")
 			c.updateFunc(ctx, e.Object)
+			c.logger.LogCtx(ctx, "level", "debug", "message", "reconciled object")
```
xh3b4sd: What about the methods do this logging on their own? We should then also have the event key present, which is missing now.

controller/controller.go
```diff
@@ -403,10 +409,15 @@ func (c *Controller) bootWithError(ctx context.Context) error {
...
 		close(c.booted)
+
+		c.logger.LogCtx(ctx, "level", "debug", "message", "processing object events")
+
 		err := c.ProcessEvents(ctx, deleteChan, updateChan, errChan)
 		if err != nil {
 			return microerror.Mask(err)
 		}
+
+		c.logger.LogCtx(ctx, "level", "debug", "message", "stopped processing object events")
```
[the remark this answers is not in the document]
kopiczko: And also makes it more aligned with what we usually do in resources.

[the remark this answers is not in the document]
xh3b4sd: Please make it one of the following, but let's not mix them. 

```
processing object events
processed object events
```

```
started processing object events
stopped processing object events
```

I would go with the first one.

controller/controller.go
xh3b4sd: > not the other way around

The resources all do it internally.

controller/controller.go
```diff
@@ -296,6 +300,8 @@ func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Ev
 		case <-ctx.Done():
 			return nil
 		}
+
+		c.logger.LogCtx(ctx, "level", "debug", "message", "reconciled event")
```
xh3b4sd: There is no `reconciling event` log.

controller/controller.go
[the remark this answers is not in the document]
kopiczko:
```suggestion
		c.logger.LogCtx(ctx, "level", "debug", "message", "processed processing object events")
```

controller/controller.go
kopiczko: This is a left over.

```suggestion
```

controller/controller.go
```diff
@@ -296,6 +300,8 @@ func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Ev
 		case <-ctx.Done():
 			return nil
 		}
+
```
kopiczko:
```suggestion
```

controller/controller.go
```diff
@@ -403,10 +407,15 @@ func (c *Controller) bootWithError(ctx context.Context) error {
...
 		close(c.booted)
+
+		c.logger.LogCtx(ctx, "level", "debug", "message", "processing object events")
+
 		err := c.ProcessEvents(ctx, deleteChan, updateChan, errChan)
 		if err != nil {
 			return microerror.Mask(err)
 		}
+
+		c.logger.LogCtx(ctx, "level", "debug", "message", "processed processing object events")
```
xh3b4sd:
```suggestion
		c.logger.LogCtx(ctx, "level", "debug", "message", "processed object events")
```

controller/controller.go
```diff
@@ -254,14 +254,16 @@ func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Ev
...
 				if err != nil {
-					c.logger.LogCtx(ctx, "level", "error", "message", fmt.Sprintf("cannot create accessor %T", e.Object), "stack", fmt.Sprintf("%#v", err))
+					c.logger.LogCtx(ctx, "level", "error", "message", fmt.Sprintf("failed to create accessor %T", e.Object), "stack", fmt.Sprintf("%#v", err))
 				} else {
 					ctx = setLoggerCtxValue(ctx, loggerKeyObject, accessor.GetSelfLink())
 					ctx = setLoggerCtxValue(ctx, loggerKeyVersion, accessor.GetResourceVersion())
 				}
 			}
 
+			c.logger.LogCtx(ctx, "level", "debug", "message", "reconciling object")
```
tuommaki: Is the information already present in the `ctx` or would it make sense to add the name / identifier of the object that is being reconciled? Also the same thing below for similar lines.

kopiczko: Batteries included.

```
D 01/24 14:11:55 processing object events | operatorkit/controller/controller.go:411 | controller=test-controller
D 01/24 14:11:55 integration-controlflow-test/nodeconfigs/test-obj reconciling object | operatorkit/controller/controller.go:287 | controller=test-controller | event=update | loop=0 | version=364
D 01/24 14:11:55 integration-controlflow-test/nodeconfigs/test-obj stop reconciliation due to finalizer added | operatorkit/controller/controller.go:351 | controller=test-controller | event=update | loop=0 | version=364
D 01/24 14:11:55 integration-controlflow-test/nodeconfigs/test-obj reconciled object | operatorkit/controller/controller.go:289 | controller=test-controller | event=update | loop=0 | version=364
```

PR #282, opened 2019-02-13
controller: handle all CRs by default

kopiczko:
When migrating operators to VOO it looks better to just drop Handles
than setting it to `func (obj interface{}) { return true }` everywhere.

Now we use label selectors instead. We should drop this feature after we
don't have version bundles anymore.


controller/controller.go
```diff
@@ -225,16 +225,19 @@ func (c *Controller) ProcessEvents(ctx context.Context, deleteChan chan watch.Ev
 		for {
```
kopiczko: Changes to this file shouldn't be in this PR. I'll revert them.

JosephSalisbury:
@kopiczko do we have tracking for dropping this in the future?

kopiczko:
I have that in my notes, but nothing in any official place.

PR #288, opened 2019-02-28
resource/configmap: add

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/5461.

This is mostly generated by the script below.
The only manual changes are in this commit: https://github.com/giantswarm/operatorkit/pull/288/commits/a4ae7bccb6056e943e1d62601d4133243d4b42a0.

```bash
replace() {
    local a=$1; shift
    local b=$1; shift
    local dir=$1; shift
    # escape backshashes
    local a="${a//\//\\/}"
    local b="${b//\//\\/}"
    for file in $(find $dir \( -path ./.git -o -path ./vendor \)  -prune -o -type f -print); do
        sed -i -e "s/${a}/${b}/g" $file
    done
}
cp ../release-operator/pkg/controller/resource/app ./resource/configmap
replace 'package app' 'package configmap' ./resource/configmap
replace 'App CR' ConfigMap ./resource/configmap
replace 'AppCR' ConfigMap ./resource/configmap
replace 'appCR' configMap ./resource/configmap
replace 'v1alpha1.App' 'v1.ConfigMap' ./resource/configmap
replace '"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"' '"k8s.io/api/core/v1"' ./resource/configmap
replace 'g8sClient.ApplicationV1alpha1().Apps' 'k8sClient.CoreV1().ConfigMaps' ./resource/configmap
goimports -w ./resource/configmap
```

### DIFF
--- a/resource/configmap/create.go
+++ b/resource/configmap/create.go
@@ -1,0 +1,60 @@
+package app
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+// ApplyCreateChange ensures the App CR is created in the k8s api.
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
+	appCRs, err := toAppCRs(createChange)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for _, appCR := range appCRs {
+		r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("creating App CR %#q in namespace %#q", appCR.Name, appCR.Namespace))
+
+		_, err = r.g8sClient.ApplicationV1alpha1().Apps(appCR.Namespace).Create(appCR)
+		if apierrors.IsAlreadyExists(err) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("already created App CR %#q in namespace %#q", appCR.Name, appCR.Namespace))
+		} else if err != nil {
+			return microerror.Mask(err)
+		} else {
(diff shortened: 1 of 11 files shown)


resource/configmap/create.go
```diff
@@ -0,0 +1,60 @@
+package configmap
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	v1 "k8s.io/api/core/v1"
```
rossf7:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

Minor but I prefer this alias.

resource/configmap/spec.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+type StateGetter interface {
+	// GetCurrentState returns a current state of the system for the given
+	// carnation of the observed Kubernetes object.
```
rossf7:
```suggestion
	// incarnation of the observed Kubernetes object.
```

Pretty sure is incarnation. Nice use of the word!

Maybe its stating the obvious but sometimes we want to create a single configmap other times its multiple. Should that be mentioned in the godoc? Could be useful for future people integrating the resource.

kopiczko: IDK. This is fine to produce single CM as a current/desired state. I can add a sentence here.

resource/configmap/create.go
kopiczko: Done. `corev1` import is used now.

resource/configmap/spec.go
kopiczko: Added this
```
	// incarnation of the observed Kubernetes object. The state consists of
	// multiple objects but it is fine to return a slice of single object
	// if only one object is managed by this resource.
```

kopiczko:
This is tested in https://github.com/giantswarm/app-operator/pull/51 by @tomahawk28. Going.

PR #295, opened 2019-03-28
k8srestconfig: fix string format

kopiczko:
giantswarm/operatorkit/client/k8srestconfig/k8s_rest_config.go:120: cannot use k8srestconfig.Config.InCluster and %!T(MISSING).KubeConfig} {invalid config error}]

### DIFF
--- a/client/k8srestconfig/k8s_rest_config.go
+++ b/client/k8srestconfig/k8s_rest_config.go
@@ -117,7 +117,7 @@ func New(config Config) (*rest.Config, error) {
 		return nil, microerror.Maskf(invalidConfigError, "cannot use %T.Address and %T.KubeConfig", config)
 	}
 	if config.InCluster && config.KubeConfig != "" {
-		return nil, microerror.Maskf(invalidConfigError, "cannot use %T.InCluster and %T.KubeConfig", config)
+		return nil, microerror.Maskf(invalidConfigError, "cannot use %T.InCluster and %T.KubeConfig", config, config)
 	}
 
 	if config.Address != "" {
--- a/client/k8srestconfig/k8s_rest_config.go
+++ b/client/k8srestconfig/k8s_rest_config.go
@@ -111,10 +111,10 @@ func New(config Config) (*rest.Config, error) {
 
 	// Settings.
 	if config.Address == "" && !config.InCluster && config.KubeConfig == "" {
-		return nil, microerror.Maskf(invalidConfigError, "%T.Address must not be empty when not using %T.InCluster or %T.KubeConfig", config)
+		return nil, microerror.Maskf(invalidConfigError, "%T.Address must not be empty when not using %T.InCluster or %T.KubeConfig", config, config, config)
 	}
 	if config.Address != "" && config.KubeConfig != "" {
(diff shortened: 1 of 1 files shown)


kopiczko:
They do but I didn't run into these brunches :D I'll fix.

client/k8srestconfig/k8s_rest_config.go
```diff
@@ -111,19 +111,19 @@ func New(config Config) (*rest.Config, error) {
 
 	// Settings.
 	if config.Address == "" && !config.InCluster && config.KubeConfig == "" {
-		return nil, microerror.Maskf(invalidConfigError, "%T.Address must not be empty when not using %T.InCluster or %T.KubeConfig", config)
+		return nil, microerror.Maskf(invalidConfigError, "%T.Address must not be empty when not using %T.InCluster or %T.KubeConfig", config, config, config)
```
TheoBrigitte: It would make sense to format the string using `fmt.Sprintf` and position argument

```suggestion
		return nil, microerror.Maskf(invalidConfigError, fmt.Sprintf("%[1]T.Address must not be empty when not using %[1]T.InCluster or %[1]T.KubeConfig", config))
```

client/k8srestconfig/k8s_rest_config.go
```diff
@@ -111,19 +111,19 @@ func New(config Config) (*rest.Config, error) {
...
 	}
 	if config.Address != "" && config.KubeConfig != "" {
-		return nil, microerror.Maskf(invalidConfigError, "cannot use %T.Address and %T.KubeConfig", config)
+		return nil, microerror.Maskf(invalidConfigError, "cannot use %T.Address and %T.KubeConfig", config, config)
```
TheoBrigitte:
```suggestion
		return nil, microerror.Maskf(invalidConfigError, fmt.Sprintf("cannot use %[1]T.Address and %[1]T.KubeConfig", config))
```

client/k8srestconfig/k8s_rest_config.go
```diff
@@ -111,19 +111,19 @@ func New(config Config) (*rest.Config, error) {
...
 	}
 	if config.InCluster && config.KubeConfig != "" {
-		return nil, microerror.Maskf(invalidConfigError, "cannot use %T.InCluster and %T.KubeConfig", config)
+		return nil, microerror.Maskf(invalidConfigError, "cannot use %T.InCluster and %T.KubeConfig", config, config)
```
TheoBrigitte:
```suggestion
		return nil, microerror.Maskf(invalidConfigError, fmt.Sprintf("cannot use %[1]T.InCluster and %[1]T.KubeConfig", config))
```

client/k8srestconfig/k8s_rest_config.go
TheoBrigitte: If you approve this change you need to add "fmt" to import.

kopiczko: I think we agreed a while ago that we don't want to use `%[n]` in our string formats. Also even we don't need fmt.Sprintf because Maskf calls it under the hood anyway. I'll go with the current version.

PR #320, opened 2019-10-16
Update subresource by update function

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/7299

SInce `Status` subresource is a part of spec, we should update it through `.Update` function. 

Already tested in `centaur/h0cg9`


tomahawk28:
Hi, @xh3b4sd, Any particular reason you switched it from `Update` to `UpdateStatus` ? https://github.com/giantswarm/operatorkit/pull/257/files#diff-70c9bcff1743248a6930e292358e2c63R129

PR #324, opened 2019-10-18
controller: make event handling concurrent

JosephSalisbury:


controller/controller.go
```diff
@@ -27,12 +30,15 @@ import (
 )
 
 const (
+	concurrency = 16
```
JosephSalisbury: This should be configurable, but I'm not certain how useful that is. Either it's lower, and users have less concurrency, or higher and users have more unused goroutines.

My hunch is to set it to a default value to begin with. I have some ideas for creating and deleting workers dynamically, so we could eventually get rid of this.

controller/controller.go
```diff
@@ -237,9 +245,80 @@ func (c *Controller) deleteFunc(ctx context.Context, obj interface{}) {
 	}
 }
 
+// hashEvent takes an event and produces an integer hash that:
+// - is guaranteed to be stable for the lifetime of an object,
+// - is in the bounds 0 <= int < c.concurrency.
+func (c *Controller) hashEvent(ctx context.Context, event watch.Event) int {
+	v := reflect.ValueOf(event.Object).Elem()
```
JosephSalisbury: Not deep enough into apimachinery to find a better way to do this. General problem is to gets some stable ID from `runtime.Object` to use as input to the hash function. Probably better ways to do it.

controller/controller.go
```diff
@@ -237,9 +245,80 @@ func (c *Controller) deleteFunc(ctx context.Context, obj interface{}) {
...
+
+	channels := make([]eventChannels, c.concurrency)
+
+	for i := 0; i < c.concurrency; i++ {
+		channels[i] = eventChannels{
+			deleteChan: make(chan watch.Event),
+			updateChan: make(chan watch.Event),
+		}
+
+		go c.processEvents(ctx, i, channels[i].deleteChan, channels[i].updateChan)
```
JosephSalisbury: I'm unsure of the context management here. When launching multiple goroutines, I'm worried that it's actually the same context being used everywhere.

JosephSalisbury:
yo @giantswarm/sig-operator - i don't reckon this is ready to merge straight away (it's a change with implications), but I'd like some feedback.

My general next steps are to try this out with a larger operator (probably aws-operator). We'd need to check general safety, and I'm also curious if there's a material speed benefit with reconciling node pool clusters, considering that they have a larger number of CRs

thanks <3

JosephSalisbury:
Don't have bandwidth to push this further, leaving for now.

PR #350, opened 2019-12-20
Allow developers to disable adding finalizers to custom resources

fiunchinho:
Learning during operator week, we found this use case where we may want to delete resources that we are watching. But since operatorkit is adding finalizers to custom resources, it can get prevent other controllers from removing the resource.

With this option, we can config whether or not we want operatorkit to add finalizers to custom resources.

Not sure if it will be useful in our dayly job, I just wanted to get deeper into operatorkit, and this was a good excuse :smile:

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -87,6 +87,8 @@ type Config struct {
 	// runtime objects the controller watches is performed. Defaults to
 	// DefaultResyncPeriod.
 	ResyncPeriod time.Duration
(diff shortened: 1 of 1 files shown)


kopiczko:
I would be also reluctant to merge this. Let's first discuss if we want keep the cleaner-operator and its design in the new year.

fiunchinho:
Closing after reading feedback!

PR #354, opened 2020-01-13
Add flag for kubeconfig file path

tuommaki:
When configuring k8sclient in operators, one may either specify path to kubectl
configuration file or fill parameters for restconfig.

### DIFF
--- a/flag/service/kubernetes/kubernetes.go
+++ b/flag/service/kubernetes/kubernetes.go
@@ -8,9 +8,10 @@ import (
 // Kubernetes is a data structure to hold Kubernetes specific command line
 // configuration flags.
 type Kubernetes struct {
-	Address    string
-	InCluster  string
-	KubeConfig string
-	TLS        tls.TLS
-	Watch      watch.Watch
+	Address        string
+	InCluster      string
+	KubeConfig     string
+	KubeConfigPath string
+	TLS            tls.TLS
+	Watch          watch.Watch
 }


tfussell:
Nice! This was annoying for me working on my operator during the hackathon. Isn't there some more wiring that has to be done or is that handled by the application using operatorkit?

tfussell:
Nevermind, I just saw your corresponding azure-operator PR.

flag/service/kubernetes/kubernetes.go
```diff
@@ -8,9 +8,10 @@ import (
...
 type Kubernetes struct {
-	Address    string
-	InCluster  string
-	KubeConfig string
-	TLS        tls.TLS
-	Watch      watch.Watch
+	Address        string
+	InCluster      string
+	KubeConfig     string
+	KubeConfigPath string
```
tfussell: It might be clearer to call this `KubeConfig` and call the other `KubeConfigData`. In most Kubernetes applications (thinking of `kubelet`), `--kubeconfig` is the path to the file rather than the file content.

tuommaki: Makes sense.

tuommaki: ...but I think there could be a bit too much disturbance at this point when many places accept the variable but fail during runtime when the string contains filename instead of kubeconfig. Those need to be mapped and then updated in one go. Sounds kind of SWH task for me. I'll got with this for now.

tuommaki: s/I'll got with this for now./I'd like to go with this for now./ :slightly_smiling_face:

tfussell: That's true. Just thought I'd mention for the future.

PR #356, opened 2020-01-15
controller: support full-fledged label selector

kopiczko:
This is a braking change but I don't think we use it in many places if at all. I need this for `prometheus-config-controller` because I want to select Services on which label is set regardless the label value. With simple equality it isn't possible.

For migration https://godoc.org/k8s.io/apimachinery/pkg/labels#SelectorFromSet can be used by replacing `MatchLabels: data` with `Selector: labels.SelectorFromSet(data)`.

I will test it with `prometheus-config-controller` and announce braking change in  `#news-dev` in Slack.


controller/types.go
```diff
@@ -0,0 +1,23 @@
+package controller
+
+// mapLabels implements labels.Labels. See
+// https://godoc.org/k8s.io/apimachinery/pkg/labels#Labels.
```
kopiczko: I couldn't find anything in the API that would convert map[string]string into labels.Labels. If you know about something or have suggestions LMK.

kopiczko: It was labels.Set.

kopiczko:
Tested on gorgoth with:

- https://github.com/giantswarm/kvm-operator/pull/806
- https://github.com/giantswarm/prometheus-config-controller/pull/139

PR #483, opened 2020-11-24
Allow filtering objects on annotation

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/14541

⚠️ This is a breaking change. If it is approved I'd release `v5.0.0` after that.


MarcelMue:
I am not against this. I don't quite get why this is necessary though. What's the use case where we have annotations but not labels?

fiunchinho:
Could we use this mechanism to filter objects based on the `pause` annotation instead of [doing it in code](https://github.com/giantswarm/operatorkit/blob/d35c2261fac41910c8dc379fd13dd28c226219f2/pkg/controller/controller.go#L470-L476)?

tfussell:
Isn't this effectively the same as the `HandlesFunc` we got rid of earlier this year in 163bc8a9de3897a094bc46e4bec38b450b2ab397? I just went through converting `kvm-operator` to use labels instead of annotations for that reason.

kopiczko:
> What's the use case where we have annotations but not labels?

Our use-case is to enable feature if the annotation exists. We want to be able to test our controller when there is a bunch of existing App CRs and most of them we don't want to touch. When it comes to feature-toggles I think annotations are the unusual way of enabling them. So we decided to go with annotations.

> Could we use this mechanism to filter objects based on the `pause` annotation instead of [doing it in code](https://github.com/giantswarm/operatorkit/blob/d35c2261fac41910c8dc379fd13dd28c226219f2/pkg/controller/controller.go#L470-L476)?

What we want is inverted pause annotation. Something that would enable reconciling instead of preventing from reconciling.

> Isn't this effectively the same as the `HandlesFunc` we got rid of earlier this year in [163bc8a](https://github.com/giantswarm/operatorkit/commit/163bc8a9de3897a094bc46e4bec38b450b2ab397)? I just went through converting `kvm-operator` to use labels instead of annotations for that reason.

I didn't know kvm-operator was using annotations to filter objects. `HandlesFunc` was originally introduced to select `ResourceSet` when we had version bundles. The purpose was different.

---

There are two solutions for us to avoid merging this:

1. Go with labels instead.
2. Have a handler that would cancel reconciliation if the annotation is missing. What I didn't like is that this handler would have to be first on the list of handlers. We in general want to avoid enforcing order of handlers. That was a trigger for me to introduce this PR.`

kopiczko:
I just figured we can use `config-controller.giantswarm.io/version` for that purpose. So not really needed. Closing this. Thank you for all your input.

PR #484, opened 2020-11-25
Add namespace in config & and wired it to manager

tomahawk28:
At some points, we need to bind the controller to a specific namespace. This change is to set the namespace from the controller generator. 

Slack conversation: https://gigantic.slack.com/archives/C2MS2VB37/p1606222861091500

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Updated
+
+- Add `namespace` into controller setting. 
+
 ## [4.0.0] - 2020-10-27
 
 ## Updated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## Updated
+## Added
 
 - Add `namespace` into controller setting. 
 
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -128,6 +130,7 @@ type Controller struct {
 	sentry                 sentry.Interface
 
 	name         string
+	namespace    string
```
tomahawk28, line 133: I'm not going to block empty string in a namespace since many operators still want to reconcile the object in all namespaces.

pkg/controller/controller.go
```diff
@@ -100,6 +100,8 @@ type Config struct {
 	// The name used should be unique in the kubernetes cluster, to ensure that
 	// two operators which handle the same resource add two distinct finalizers.
 	Name string
+	// Namespace is where the controller would reconcile the runtime objects.
```
kopiczko, line 103:
```suggestion
	// Namespace is where the controller would reconcile the runtime objects.
	// Empty string means all namespaces.	
```

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Updated
```
kopiczko, line 12:
```suggestion
## Added
```

kopiczko (approved):

rossf7 (commented):
Looks good but CI is not happy. 

Seems to be some weird imports thing but let's be sure the integration tests pass.

tomahawk28:
It was whitespaces, but not so sure why CI think it was go imports issue. I just fixed it now :)

rossf7 (approved):
LGTM

PR #498, opened 2020-12-18
Change `operatorkit_controller_error_total` metric

tuommaki:
Change `operatorkit_controller_error_total` from gauge to counter and
provide labels for CR name and when available, cluster ID and release
version. Third party errors caught by `controller-runtime` error handler
are not counted anymore.

## Checklist

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -307,7 +307,7 @@ func (c *Controller) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 	if err != nil {
 		// Microerror creates an error event on the object when kind and description is set.
 		c.event.Emit(ctx, obj, err)
-		errorGauge.Inc()
+		errorCounterVec.With(errorMetricLabels(obj)).Inc()
 		c.sentry.Capture(ctx, err)
 		c.logger.Errorf(ctx, err, "failed to reconcile")
 		return reconcile.Result{}, nil
@@ -329,7 +329,6 @@ func (c *Controller) bootWithError(ctx context.Context) error {
 		for {
 			resetWait := c.resyncPeriod * 4
 			time.Sleep(resetWait)
-			errorGauge.Set(0)
 		}
 	}()
 
@@ -348,7 +347,6 @@ func (c *Controller) bootWithError(ctx context.Context) error {
 					return
 				}
 
-				errorGauge.Inc()
 				c.logger.Errorf(ctx, err, "caught third party runtime error")
 			},
 		}
--- a/pkg/controller/metric.go
+++ b/pkg/controller/metric.go
@@ -19,7 +19,7 @@ var (
 			Name:      "error_total",
 			Help:      "Number of reconciliation errors.",
 		},
-		[]string{"cr_name", "cluster_id", "release_version"},
+		[]string{"cr_name", "cr_kind", "cluster_id", "release_version"},
 	)
 	eventHistogram = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
--- a/pkg/controller/metric.go
+++ b/pkg/controller/metric.go
@@ -2,6 +2,8 @@ package controller
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 const (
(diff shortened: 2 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -329,7 +329,6 @@ func (c *Controller) bootWithError(ctx context.Context) error {
 		for {
 			resetWait := c.resyncPeriod * 4
 			time.Sleep(resetWait)
-			errorGauge.Set(0)
```
tuommaki, line 332: We have alerts with `rate()` so there's no need to set it to `0`. With labels this also doesn't make sense anymore.

pkg/controller/controller.go
```diff
@@ -348,7 +347,6 @@ func (c *Controller) bootWithError(ctx context.Context) error {
 					return
 				}
 
-				errorGauge.Inc()
```
tuommaki, line 351: Without context information, this count doesn't add much. While here, I was wondering if we should gracefully exit at this point? In case there's third party error that's not related to reconciled CR, I'd try restarting the operator and if that fails (i.e. operator gets into restart loop) we'd get paged for that. In case of an error here, I believe there's no feasible way to continue anyway?

pkg/controller/metric.go
```diff
@@ -10,13 +12,14 @@ const (
 )
 
 var (
-	errorGauge = prometheus.NewGauge(
-		prometheus.GaugeOpts{
+	errorCounterVec = prometheus.NewCounterVec(
```
tuommaki, line 15: Counter makes more sense for error counting and we already have alerts using `rate()`.

pkg/controller/metric.go
```diff
@@ -10,13 +12,14 @@ const (
...
+		prometheus.CounterOpts{
 			Namespace: PrometheusNamespace,
 			Subsystem: PrometheusSubsystem,
 			Name:      "error_total",
 			Help:      "Number of reconciliation errors.",
 		},
+		[]string{"cr_name", "cluster_id", "release_version"},
```
tuommaki, line 22: Provide insight into metric from context where the error occurred.

pkg/controller/controller.go
fiunchinho: If we don't increment any metric from the error handler, we might as well remove it or is it doing anything that I'm missing?

tuommaki: Well, it's logging, but my thinking here was that it might actually make sense to `os.Exit(1)` after logging.

pkg/controller/metric.go
```diff
@@ -30,6 +33,39 @@ var (
...
+		labels["cluster_id"] = ""
+	}
+
+	if v, exists := m.GetLabels()["release.giantswarm.io/version"]; exists {
+		labels["release_version"] = v
+	} else {
+		labels["release_version"] = ""
+	}
+
+	return labels
```
kopiczko, line 70: I have mixed feelings when it comes to those labels. They are GS specific (maybe besides cluster api ones) which is IMO wrong direction for a library. They also increase metrics cardinality greatly since cluster IDs and CR names are unbound. Also `cr_`. Not only CRs are reconciled. We sometimes are reconciling some builtin objects.

tuommaki: I understand that, but on the other hand the error metric is pretty much useless without those labels and they cannot be injected outside of `operatorkit`. I'm not worried about cardinality at all since these metrics are only emitted on errors.

tuommaki: How about labels: `object_kind`, `object_name` and dropping `cluster_id` and `release_version` ? Then it's not bound to us nor CR concept and in Cluster API the CR name is currently equal with Cluster ID.

fiunchinho: That's not the case for all the CRs, like `MachinePool`.

tuommaki: Oh, that's true.

kopiczko: `object_kind`, `object_name`, `object_namespace` would be fine with me in terms of being generic. In terms of cardinality I'm not convinced. If something reconciling pods is deployed with a bug then we can end up with thousands of additional metrics.

But let's focus on a problem. Why having those labels would make things useful? If it is specific problem in your team maybe writing a handler wrapper outside operatorkit that would emit metrics you like is an option?

tuommaki:
> Why having those labels would make things useful?

These days operators reconcile relatively big number of CRs and especially with provider operators and Cluster API, different workload clusters are probably in different subscriptions so understanding which CR caused the unexpected error helps a lot, especially if it's something that we can't do about (so we can silence that specific cluster's errors).

> If it is specific problem in your team maybe writing a handler wrapper outside operatorkit that would emit metrics you like is an option?

It's not specific only to our team, but after recent discussions I believe that writing another layer on top of operatorkit is the best direction for now so I'll drop this.

PR #501, opened 2020-12-23
Adding last reconciled metrics

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/14847 

It's about adding `operatorkit_controller_last_reconciled` gauge metric into operatorkit, so we can set the current timestamp for each successful reconciliation and later utilize it for replacing the alerts based in `up` metrics. (e.g.  `ChartOperatorDown`) 

## Checklist

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -315,8 +315,6 @@ func (c *Controller) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		return reconcile.Result{}, nil
 	}
 
-	reportLastReconciled(obj)
-
 	return res, nil
 }
 
@@ -501,6 +499,7 @@ func (c *Controller) reconcile(ctx context.Context, req reconcile.Request, obj i
 		}
 
 		t.ObserveDuration()
+		unsetLastReconciled(obj)
 	} else {
 		event := "update"
 
@@ -515,6 +514,7 @@ func (c *Controller) reconcile(ctx context.Context, req reconcile.Request, obj i
 		}
 
 		t.ObserveDuration()
+		reportLastReconciled(obj)
 	}
 
 	return reconcile.Result{}, nil
@@ -610,26 +610,44 @@ func unsetLoggerCtxValue(ctx context.Context, key string) context.Context {
 	return ctx
 }
 
+func unsetLastReconciled(o interface{}) {
+	labels := getLabels(o)
+	if labels == nil {
+		return
+	}
+
+	lastReconciledGauge.DeleteLabelValues(labels...)
+}
+
 func reportLastReconciled(o interface{}) {
+	labels := getLabels(o)
+	if labels == nil {
+		return
+	}
+
+	lastReconciledGauge.WithLabelValues(labels...).SetToCurrentTime()
+}
+
+func getLabels(o interface{}) []string {
 	var kind string
 	{
 		obj, ok := o.(pkgruntime.Object)
 		if !ok {
-			return
+			return nil
 		}
 
 		gvks, _, err := scheme.Scheme.ObjectKinds(obj)
 		if pkgruntime.IsNotRegisteredError(err) {
 			gvks, _, err = apiextensionsscheme.Scheme.ObjectKinds(obj)
 			if err != nil {
-				return
+				return nil
 			}
 		} else if err != nil {
-			return
+			return nil
 		}
 
 		if len(gvks) == 0 {
-			return
+			return nil
 		}
 
 		kind = gvks[0].Kind
@@ -639,16 +657,12 @@ func reportLastReconciled(o interface{}) {
 	{
 		obj, err := meta.Accessor(o)
 		if err != nil {
-			return
+			return nil
 		}
 
 		name = obj.GetName()
 		namespace = obj.GetNamespace()
 	}
 
-	lastReconciledGauge.WithLabelValues(
-		kind,
-		name,
-		namespace,
-	).SetToCurrentTime()
+	return []string{kind, name, namespace}
 }
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -315,7 +315,7 @@ func (c *Controller) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		return reconcile.Result{}, nil
 	}
 
-	reportLastReconciled(obj)
+	reportLastReconciled(obj, c.name)
 
 	return res, nil
 }
@@ -606,7 +606,7 @@ func unsetLoggerCtxValue(ctx context.Context, key string) context.Context {
 	return ctx
 }
 
-func reportLastReconciled(o interface{}) {
+func reportLastReconciled(o interface{}, controllerName string) {
 	var kind string
 	{
 		obj, ok := o.(pkgruntime.Object)
@@ -633,5 +633,6 @@ func reportLastReconciled(o interface{}) {
 
 	lastReconciledGauge.WithLabelValues(
 		kind,
+		controllerName,
 	).SetToCurrentTime()
 }
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -11,7 +11,6 @@ import (
 	"syscall"
 	"time"
 
-	apiextensionsscheme "github.com/giantswarm/apiextensions/v3/pkg/clientset/versioned/scheme"
 	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/k8sclient/v5/pkg/k8sclient"
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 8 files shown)


pkg/controller/controller.go
```diff
@@ -605,3 +607,41 @@ func unsetLoggerCtxValue(ctx context.Context, key string) context.Context {
...
+
+func reportLastReconciled(o interface{}) {
+	var kind string
+	{
+		obj, ok := o.(pkgruntime.Object)
+		if !ok {
+			return
+		}
+
+		gvks, _, err := scheme.Scheme.ObjectKinds(obj)
```
tomahawk28, line 619: Actually, If I could use `meta.TypeAccessor` (https://godoc.org/k8s.io/apimachinery/pkg/api/meta#TypeAccessor), it would be much easier to fetch `kind` information, but actually all `TypeMeta` is empty in default so I need to use this `ObjectKinds` instead.

pkg/controller/controller.go
```diff
@@ -312,6 +313,7 @@ func (c *Controller) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		c.logger.Errorf(ctx, err, "failed to reconcile")
 		return reconcile.Result{}, nil
 	}
+	reportLastReconciled(obj)
```
tomahawk28, line 316: Operators would report `last_reconciled` metrics only if they finished reconciliation successfully. That's why I don't like to use `defer` here.

pkg/controller/controller.go
```diff
@@ -605,3 +608,46 @@ func unsetLoggerCtxValue(ctx context.Context, key string) context.Context {
...
+func reportLastReconciled(o interface{}) {
+	var kind string
+	{
+		obj, ok := o.(pkgruntime.Object)
+		if !ok {
+			return
+		}
+
+		gvks, _, err := scheme.Scheme.ObjectKinds(obj)
+		if pkgruntime.IsNotRegisteredError(err) {
```
tomahawk28, line 621: If this object is not registered in k8s upstream runtime, probably we should use another runtime package in `apiextensions` repository.

tomahawk28:
Tested in `ginger` you can see the graph as below. 
image

pkg/controller/controller.go
```diff
@@ -605,3 +609,46 @@ func unsetLoggerCtxValue(ctx context.Context, key string) context.Context {
 
 	return ctx
 }
+
+func reportLastReconciled(o interface{}) {
+	var kind string
+	{
+		obj, ok := o.(pkgruntime.Object)
+		if !ok {
+			return
```
tomahawk28, line 618: I don't think it's worth printing errors or return an error for metrics since it's not a major business of the controller. WDYT?

rossf7: What happens if the metrics are not emitted? Will the alerts based on the last reconciled timestamp still fire?

If so then I think thats OK. As long as it's clear how to debug this.

pkg/controller/controller.go
rossf7: For my interest do you know if the controller runtime client sets this type information?

It's a large change but I think in sig-operator we wanted to move away from the generated client to the controller client. This might be another reason for that change.

pkg/controller/metric.go
```diff
@@ -27,9 +27,19 @@ var (
 		},
 		[]string{"event"},
 	)
+	lastReconciledGauge = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: PrometheusNamespace,
+			Subsystem: PrometheusSubsystem,
+			Name:      "last_reconciled",
+			Help:      "",
```
rossf7, line 35: Please add some help text to explain what this metric is for.

rossf7 (commented):
Approach looks good. Just a question on the error handling and the metric needs some help text.

pkg/controller/metric.go
```diff
@@ -27,9 +27,19 @@ var (
...
+		},
+		[]string{"kind", "name", "namespace"},
```
kopiczko, line 37: This is another proposition to have metric with an unbound label cardinality. This disturbs me.

But if @giantswarm/sig-monitoring and @giantswarm/team-atlas are ok with that I'm fine too.

pkg/controller/controller.go
tomahawk28: operatorkit collect another timestamp metric and it is using controller runtime, and I see they have no problem with type information. https://github.com/giantswarm/operatorkit/blob/dc6f5aa4fb98c5cc3b42bfada11793e41144cbdd/pkg/controller/collector/timestamp.go#L60-L87

So you're right at this point; this is another reason we need to move to controller runtime client.

pkg/controller/controller.go
tomahawk28: >  Will the alerts based on the last reconciled timestamp still fire?

Yes, exactly. Alerts would be based on the last reconciled timestamp.

pkg/controller/controller.go
rossf7: Cool I'll find the SIG issue and post there. I'll also create an issue in the bat board.

pkg/controller/metric.go
rossf7: Thanks @kopiczko good to raise this. We're getting more CRs in the control planes. So this could be an issue in busy CPs with lots of TCs. I guess app CRs and certconfig CRs are the most numerous.

@tomahawk28 Let's be sure Atlas and the SIG are fine with this.

tomahawk28: This would create lots of metrics labels by considering the number of apps CRs/chart CRs in each CP. (appx. 60 for each)
I would like to know whether we have any limitations or guidelines limiting the number of label values. 

One thing might we can considering is that, turn these alerts off if we use operatorkit for general k8s objects, likely secret, service objects. Since this could be an hour-long discussion, I would move it to the sig-monitoring channel.

JosephSalisbury: Could we get away with just having the kind? So, we'd have one metric per operator / controller. I can't personally think of a situation where we'd only reconcile _some_ CRs

tomahawk28: Thanks @JosephSalisbury,  In the case of `App`, `Charts` CRD, and some provider-specific CRDs, they are reconciled to the operators with matching versions. Because of this, only `kind` value is not enough to see the reconciliation problem. 
do we have any controller-specific value which we could identify the matching labels? likely selector?

JosephSalisbury: I don't think I follow why only having the `kind` label isn't sufficient - from https://github.com/giantswarm/operatorkit/pull/501#issuecomment-753994356, it looks like all custom resources have very similar timestamps.

tomahawk28: I thought that we need a fingerprint for each controller, for example, cluster-operator-0-23-16 and cluster-operator-0-23-17 would be reconciled to the same kind type but totally different matching labels, right?

See belows for labels. 
```
✔{app="app-operator",app_instance="app-operator-unique",app_name="app-operator",app_version="0.0.0",cluster_id="ginger",cluster_type="host",exported_namespace="giantswarm",instance="100.64.5.144:8000",job="host-cluster-ginger-workload",kind="App",name="vertical-pod-autoscaler-app-unique",namespace="giantswarm",node="ip-10-0-5-53.eu-west-1.compute.internal",pod_phase="Running",pod_ready="true",provider="aws"}

```
But now I also realized maybe we could utilize `app_version` and `app_name` to let us know the location of operators. 
so having a `kind` label only is also an option, WDYT?

JosephSalisbury: yeah, different operators will have different `instance` labels, so they'll be differentiable. if you're happy with it, i think it lets us sidestep the whole unbounded cardinality issue

kopiczko: From the title of https://github.com/giantswarm/giantswarm/issues/14847 it sounds to me like the only labels that are really needed are:

- operator name
- operator version
- controller name

Based on that you can clearly tell which controller in which operator in which version is not reconciling.

tomahawk28: Sure, I have `name`, `app_version`, and `kind` as labels, I don't think the controller name is worth it since `kind` itself already represents such information.

tomahawk28:
Tested on ginger; now metrics have only `kind` labels. 
image

pkg/controller/metric.go
kopiczko: There can be a situation in which 2 separate controllers reconcile the same kind. Maybe it isn't very useful but possible. BTW that's going to happen in config-controller (as long as we don't have AppCatalogEntry CRs for test catalogs).

tomahawk28: If the case that multiple controllers are reconciled into the same kind of object is prevalent, then I think adding the controller name into the label makes sense. thanks to let me know that.

tomahawk28: I tested with an additional label with controller name, but found out most of the cases, we are assinging `project.Name()` to the name of the controller. 
https://github.com/giantswarm/app-operator/blob/5e081b766dbee2e14cb3f5869946f97302f0b9ea/service/controller/app/app.go#L98-L115
@kopiczko I think that's something we should modify from the operator itself, not operatorkit should do, WDYT?
PTAL.

kopiczko: > but found out most of the cases, we are assinging `project.Name()` to the name of the controller.

That's ok.

kopiczko: I'd still remove kind label but I won't block because of that. I think what controller is watching is irrelevant from the alert standpoint.

kopiczko (approved):

pkg/controller/metric.go
tomahawk28: As we have a controller name as label, so now I agree we don't need kind information. I will delete that we can introduce later if we really need that.

rossf7 (approved):

PR #512, opened 2021-01-28
Improve timestamp collector

tfussell:
We recently found in kvm-operator that the timestamp collector was allocating a huge number of objects (~10MB/s) and it seems to be due to listing all pods in all namespaces (one controller watches pods) and then filtering client-side. This PR switches to server-side filtering so we only get the objects we're watching from the server. I will test today or Monday how this affects memory usage, but I think it's a useful improvement either way.

Testing in:
- https://github.com/giantswarm/kvm-operator/pull/1136
- https://github.com/giantswarm/node-operator/pull/181
- https://github.com/giantswarm/config-controller/pull/82

## Checklist


pkg/controller/internal/selector/selector.go
```diff
@@ -1,46 +0,0 @@
-package selector
```
tfussell, line 1: It became useful to use the upstream Kubernetes apimachinery labels package over this custom interface so that we could pass a selector from a library consumer to a controller to the collector to the ctrlClient selector without conversions. It makes it a breaking change so I will bump the major version if this removal is acceptable. An example of a PR converting an operator to use this new format can be found here https://github.com/giantswarm/node-operator/pull/181

pkg/controller/collector/timestamp.go
```diff
@@ -57,43 +65,41 @@ func NewTimestamp(config TimestampConfig) (*Timestamp, error) {
...
+	})
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
 	for _, object := range list.Items {
-		if !t.selector.Matches(selector.NewLabels(object.GetLabels())) {
-			// Skip objects that do not match the selector labels
-			continue
-		}
```
tfussell, line 78: Here are the most important changes, passing in selectors for server-side filtering rather than doing it client-side.

pkg/controller/collector/timestamp.go
```diff
@@ -57,43 +65,41 @@ func NewTimestamp(config TimestampConfig) (*Timestamp, error) {
 }
 
 func (t *Timestamp) Collect(ch chan<- prometheus.Metric) error {
-	ctx := context.Background()
-	gvk, err := apiutil.GVKForObject(t.newRuntimeObjectFunc(), t.k8sClient.Scheme())
-	if err != nil {
-		return microerror.Mask(err)
+	list := &metav1.PartialObjectMetadataList{}
```
tfussell, line 68: This type allows us to only get metadata rather than the whole object, saving memory and network traffic.

pkg/controller/collector/timestamp_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	pkgruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func Test_Timestamp(t *testing.T) {
```
tfussell, line 21: This isn't the prettiest unit test, but I can't come up with a better way.

pkg/controller/controller.go
```diff
@@ -87,6 +86,15 @@ type Config struct {
...
 	NewRuntimeObjectFunc func() pkgruntime.Object
+	// NewRuntimeObjectListFunc returns a new initialized pointer of a type
+	// implementing the runtime object interface. The object returned is used with
+	// the timestamp collector to fetch the list of watched object. See the example below.
+	//
+	//     func() pkgruntime.Object {
+	//        return new(corev1.ConfigMapList)
+	//     }
+	//
+	NewRuntimeObjectListFunc func() pkgruntime.Object
```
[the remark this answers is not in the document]
kopiczko: Was it wrong in general

pkg/controller/controller.go
```diff
@@ -395,10 +397,10 @@ func (c *Controller) bootWithError(ctx context.Context) error {
...
 			}).
 			WithEventFilter(predicate.Funcs{
-				CreateFunc:  func(e event.CreateEvent) bool { return c.selector.Matches(selector.NewLabels(e.Meta.GetLabels())) },
-				DeleteFunc:  func(e event.DeleteEvent) bool { return c.selector.Matches(selector.NewLabels(e.Meta.GetLabels())) },
-				UpdateFunc:  func(e event.UpdateEvent) bool { return c.selector.Matches(selector.NewLabels(e.MetaNew.GetLabels())) },
-				GenericFunc: func(e event.GenericEvent) bool { return c.selector.Matches(selector.NewLabels(e.Meta.GetLabels())) },
+				CreateFunc:  func(e event.CreateEvent) bool { return c.selector.Matches(labels.Set(e.Meta.GetLabels())) },
+				DeleteFunc:  func(e event.DeleteEvent) bool { return c.selector.Matches(labels.Set(e.Meta.GetLabels())) },
+				UpdateFunc:  func(e event.UpdateEvent) bool { return c.selector.Matches(labels.Set(e.MetaNew.GetLabels())) },
+				GenericFunc: func(e event.GenericEvent) bool { return c.selector.Matches(labels.Set(e.Meta.GetLabels())) },
```
kopiczko, line 403: This doesn't look like server-side filtering to me. Am I missing something?

pkg/controller/controller.go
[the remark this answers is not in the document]
tfussell: I don't think I've ever seen it be wrong. I guess I've talked myself into reverting this or at least making it optional.

pkg/controller/controller.go
yulianedyalkova: I believe [this](https://github.com/giantswarm/operatorkit/blob/047b5dd3303e00d6d0c772121ae8841bea3d1ee4/pkg/controller/collector/timestamp.go#L79) is the line that does the magic.

yulianedyalkova (approved):

pkg/controller/controller.go
tfussell: Oh yeah, the part about the server-side filtering is only relevant to the timestamp collector linked by Yulia. This `WithEventFilter` part was a secondary change related to refactoring the selectors/labels to keep everything consistent. It shouldn't have any effect on watches.

pkg/controller/collector/timestamp.go
```diff
@@ -57,43 +66,42 @@ func NewTimestamp(config TimestampConfig) (*Timestamp, error) {
...
+	{
+		gvk, err := apiutil.GVKForObject(t.newRuntimeObjectFunc(), t.scheme)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		gvk.Kind = fmt.Sprintf("%sList", gvk.Kind)
+		list.SetGroupVersionKind(gvk)
```
kopiczko, line 76: Are you sure you need to do all of that? I think `ctrlClient.List` will do that for you.

pkg/controller/controller.go
kopiczko: I took me a while to check this. When initially implementing this part I designed the Selector interface in a way that k8s.io/apimachinery/pkg/labels.Selector fulfils that interface (I extracted only the Matches method) but I also wanted to be allow writing a simple struct that fulfils this interface. So you can pass it without making breaking changes. So I believe you can drop all changes to `controller.go` and your tests will still compile and be green.

tfussell: Interesting. I have trouble with go interfaces sometimes. I'll give it a try.

pkg/controller/collector/timestamp.go
tfussell: I will investigate.

tfussell: Reading through the code, the client does strip off the `List` suffix from the `Kind` so there's no reason for line 75. I think we still need to to call `SetGroupVersionKind` because otherwise we'll get all objects that match the selector instead of just the kind we're interested in. I'll push a small fix taking this into account.

tfussell: Done.

pkg/controller/controller.go
tfussell: The problem is that we pass this selector into [ListOptions's LabelSelector which accepts an object satisfying k8s.io/apimachinery/pkg/labels.Selector interface](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.2/pkg/client/options.go#L321-L323) and the `Selector` from `operatorkit` doesn't define all of the required methods so we need to use `Selector` from `apimachinery` throughout (my approach so far which requires more work for consumers) or make the `Selector` from `operatorkit` define the same interface as the one from `apimachinery` (less work for consumers, more work in `operatorkit`). What do you think @kopiczko?

pkg/controller/collector/timestamp.go
tfussell: Actually, without that line I get an error saying `non-list type *v1.PartialObjectMetadataList (kind \"/v1, Kind=Pod\") passed as output` so I can't see any way to change the code here.

pkg/controller/controller.go
kopiczko: I see. Then it doesn't work. I leave the decision for you.

pkg/controller/collector/timestamp.go
kopiczko: Okay yeah this is dynamic stuff.

PR #520, opened 2021-02-10
fix double channel close panic

tfussell:
I had some panics recently because of double-closing channels. I think this resolves the problem.

## Checklist

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -130,6 +130,7 @@ type Controller struct {
 	backOffFactory         func() backoff.Interface
 	bootOnce               sync.Once
 	booted                 chan struct{}
+	stopOnce               sync.Once
 	stopped                chan struct{}
 	collector              *collector.Set
 	loop                   int64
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -283,8 +284,10 @@ func NewSelector(matchesFunc func(labels labels.Labels) bool) selector.Selector
 }
 
 func (c *Controller) Stop(ctx context.Context) {
-	c.collector.Stop(ctx)
-	close(c.stopped)
+	c.stopOnce.Do(func() {
+		c.collector.Stop(ctx)
+		close(c.stopped)
+	})
```
kopiczko, line 290: I'm fine with that but are you sure closing it multiple times isn't a bug in your code?

kopiczko (approved):

pkg/controller/controller.go
tfussell: The problem is that we can shut down the controller without exiting the rest of the program. Then if we send a SIGTERM later it will try to close the channel again.

PR #528, opened 2021-02-24
Switch back to UnstructuredList

yulianedyalkova:
https://github.com/giantswarm/kvm-operator/pull/1136#issuecomment-785149679

### DIFF
--- a/pkg/controller/collector/timestamp.go
+++ b/pkg/controller/collector/timestamp.go
@@ -2,7 +2,6 @@ package collector
 
 import (
 	"context"
-	"errors"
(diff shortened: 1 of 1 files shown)


tfussell:
This means more allocations, but should still be a big improvement with the server-side filtering.

tfussell (approved):

yulianedyalkova:
I know but I couldn't figure out why we're missing the kind in the metadata - we obviously have it in the object itself cause we're getting it when getting the UnstructuctedList. And I think it's a pretty important part of the metric we emit.

PR #532, opened 2021-03-16
Add AllowedLabels to configmap resource

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/14163

Extends the generic configmap resource to configure a list of allowed labels.

This is needed in the clusterconfigmap in cluster-operator because the configmap includes the label below which is added by app-operator so it can watch for changes.

```
labels:
  app-operator.giantswarm.io/watching: "true"
```

## Checklist

### DIFF
--- a/pkg/resource/k8s/configmapresource/new_config_map_to_update.go
+++ b/pkg/resource/k8s/configmapresource/new_config_map_to_update.go
@@ -1,8 +1,10 @@
 package configmapresource
 
 import (
+	"fmt"
 	"reflect"
 
+	"github.com/google/go-cmp/cmp"
 	corev1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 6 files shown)


rossf7:
Tested on giraffe.

pkg/resource/k8s/configmapresource/new_config_map_to_update.go
```diff
@@ -23,6 +23,19 @@ func newConfigMapToUpdate(current, desired *corev1.ConfigMap) *corev1.ConfigMap
...
 	merged.Labels = desired.Labels
 
+	if allowedLabels != nil {
+		for k, v := range current.Labels {
+			if _, exist := desired.Labels[k]; exist {
+				// If label is already in desired spec, skip it.
+				continue
+			}
+
+			if _, ok := allowedLabels[k]; ok {
```
tomahawk28, line 33: I think we can make it simple, 
```suggestion
			if allowedLabels[k] {
```

rossf7: Thanks, I based this on the app resource but this is cleaner.

tomahawk28 (approved):

PR #538, opened 2021-04-06
Remove usage of self link for Kubernetes 1.20 support

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/16295

Removes usage of self link as its disabled by default in k8s 1.20.

Tested with https://github.com/giantswarm/chart-operator/pull/736

## Checklist


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  architect: giantswarm/architect@0.17.0
+  architect: giantswarm/architect@dev:k8s-1-20
```
rossf7, line 4: Test with K8s 1.20.2 in the integration tests.

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 
 replace (
 	github.com/coreos/etcd => github.com/coreos/etcd v3.3.24+incompatible
+	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
```
rossf7, line 24: Fix nancy which flags vuln in 1.3.1.

pkg/controller/finalizer.go
```diff
@@ -91,12 +92,12 @@ func (c *Controller) hasFinalizer(ctx context.Context, obj interface{}) (bool, e
 		return false, microerror.Mask(err)
 	}
 	finalizerName := getFinalizerName(c.name)
-	selfLink := accessor.GetSelfLink()
+	uid := string(accessor.GetUID())
```
rossf7, line 95: We now use UID as the cache key. It doesn't have a string function so just casting it.

https://github.com/kubernetes/apimachinery/blob/v0.20.5/pkg/types/uid.go#L19-L22

pkg/controller/controller.go
```diff
@@ -501,7 +501,7 @@ func (c *Controller) reconcile(ctx context.Context, req reconcile.Request, obj i
 		}
 	}
 
-	ctx = setLoggerCtxValue(ctx, loggerKeyObject, m.GetSelfLink())
+	ctx = setLoggerCtxValue(ctx, loggerKeyObject, fmt.Sprintf("%s/%s", m.GetNamespace(), m.GetName()))
```
rossf7, line 504: We lose the type information here which is a shame but its also less verbose in the logs.

rossf7 (commented):
Testing on ginger.

fiunchinho (approved):
LGTM

kopiczko (approved):

PR #594, opened 2022-01-03
add boot log line

TheoBrigitte:


rossf7 (approved):

pkg/controller/controller.go
```diff
@@ -250,6 +250,8 @@ func (c *Controller) Boot(ctx context.Context) {
 	ctx = setLoggerCtxValue(ctx, loggerKeyController, c.name)
 
 	c.bootOnce.Do(func() {
+		c.logger.LogCtx(ctx, "level", "debug", "message", "booting controller")
```
tfussell, line 253:
```suggestion
		c.logger.Debugf(ctx, "booting controller")
```

pkg/controller/controller.go
```diff
@@ -267,6 +269,8 @@ func (c *Controller) Boot(ctx context.Context) {
 			c.logger.Errorf(ctx, err, "stop controller boot retries due to too many errors")
 			os.Exit(1)
 		}
+
+		c.logger.LogCtx(ctx, "level", "debug", "message", "booted controller")
```
tfussell, line 273:
```suggestion
		c.logger.Debugf(ctx, "booted controller")
```

pkg/controller/controller.go
tfussell: Using https://github.com/giantswarm/micrologger/pull/60

PR #653, opened 2023-11-09
Release v7.2.0

taylorbot:


pkg/controller/controller.go
```diff
@@ -374,12 +376,21 @@ func (c *Controller) bootWithError(ctx context.Context) error {
...
-			// ourselves.
-			MetricsBindAddress: DisableMetricsServing,
-			Namespace:          c.namespace,
-			SyncPeriod:         to.DurationP(c.resyncPeriod),
+			Cache: cacheOptions,
+			Metrics: server.Options{
+				// MetricsBindAddress is set to 0 in order to disable it. We do this
+				// ourselves.
+				BindAddress: DisableMetricsServing,
+			},
```
TheoBrigitte, line 393: What is this for? It should be mentioned in CHANGELOG, and could also be moved to a different PR unless its fixing some issue introduced by dependency upgrade.

QuentinBisson: Yes it's only fixing issues with breaking changes in controller-runtime upgrade. Behavior is the same or at least should bé from what I read upstream

pkg/controller/controller.go
```diff
@@ -397,7 +408,6 @@ func (c *Controller) bootWithError(ctx context.Context) error {
 			For(c.newRuntimeObjectFunc()).
 			WithOptions(controller.Options{
 				MaxConcurrentReconciles: 1,
-				Reconciler:              c,
```
QuentinBisson, line 400: Needed because latest ctrl runtime packages added validation and the reconcilier is added later on

pkg/controller/controller.go
QuentinBisson: See https://github.com/kubernetes-sigs/controller-runtime/commit/7ad4b4c318f8628bcab1c2b436e86f9dbd91d852 and this https://github.com/kubernetes-sigs/controller-runtime/commit/0377e0f7a41bd9d88d43811192309b193c9cd793

QuentinBisson: @TheoBrigitte I'll merge when you approve :)

QuentinBisson (approved):

